PR #1, opened 2024-08-12
Init InplaceVector

Hels15:
- Added basic container
- Basic test cases(not complete, quite lacking)


Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
...
+      std::destroy(begin + n, this->end());
+    } else {
+      const iterator end = this->end();
+      std::fill(this->begin(), end, u);
+      base::uninitialized_fill(end, end + diff, u);
+    }
+    this->change_size(diff);
+  }; // freestanding-deleted
+  constexpr void assign(std::initializer_list<T> il) {
+    if constexpr (Capacity == 0) {
```
JeffGarland, line 357: Is this correct for an empty initializer list?  It would prevent the construction of a zero length inplace vector which I think we want to allow?

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
...
+      assert(il.list() && "Cannot assign to inplace_vector with zero capacity");
+      return;
+    }
+    const auto diff = static_cast<std::ptrdiff_t>(il.size() - this->size());
+    if (diff < 0) {
+      const iterator new_end = std::copy(il.begin(), il.end(), this->begin());
```
JeffGarland, line 363: Since we're in c++20 mode here std::ranges::copy is available.

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
...
+    return pos;
+  }; // freestanding-deleted
+  constexpr iterator insert(const_iterator position, const T &x) {
+    return emplace(position, x);
+  }; // freestanding-deleted
+  constexpr iterator insert(const_iterator position, T &&x) {
+    return emplace(position, std::move(x));
+  }; // freestanding-deleted
+  constexpr iterator insert(const_iterator position, size_type n, const T &x) {
+    const iterator pos = (iterator)position;
```
JeffGarland, line 579: const_cast to be more specific?

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
...
+    for (; first != last; ++first) {
+      emplace_back(*first);
+    }
+    const iterator pos = position;
+    std::rotate(pos, old_end, this->end());
+    return pos;
+  } // freestanding-deleted
+  constexpr iterator insert(const_iterator position,
+                            std::initializer_list<T> il) {
+    const iterator pos = (iterator)position;
```
JeffGarland, line 660: Seel line 679 -- if that one converts why is the cast needed?

src/Beman/InplaceVector/tests/InplaceVector.test.cpp
```diff
@@ -0,0 +1,88 @@
+#include <Beman/InplaceVector/InplaceVector.hpp>
+
+using namespace beman::InplaceVector;
+
```
JeffGarland, line 4: this is a good start, but obviously there's plenty more cases to cover here.  Since there was a lot of controversy about constexpr would certainly be good to have cases here -- and also for the zero capacity case.

steve-downey:
On Mon, Aug 19, 2024, 11:16 Jeff Garland ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In Include/Beman/InplaceVector/InplaceVector.hpp
> <https://github.com/beman-project/InplaceVector/pull/1#discussion_r1721958977>
> :
>
> > +      return;
> +    }
> +    const auto diff = static_cast<std::ptrdiff_t>(il.size() - this->size());
> +    if (diff < 0) {
> +      const iterator new_end = std::copy(il.begin(), il.end(), this->begin());
> +      std::destroy(new_end, this->end);
> +    } else {
> +      std::copy(il.begin(), il.begin() + this->size(), this->begin());
> +      base::uninitialized_copy(il.begin() + this->size(), il.end(),
> +                               this->end());
> +    }
> +    this->change_size(diff);
> +  }; // freestanding-deleted
> +
> +  // [containers.sequences.inplace.vector.access], element access
> +  constexpr reference at(const size_type count) {
>
> https://eel.is/c++draft/inplace.vector.overview standard doesn't have
> const in the signature -- that might fall into implementer freedom though...
>

const on size_t as a parameter is a no-op though?  It does mean that the
implementation can't modify their copy of the parameter, but the caller
doesn't care.  (Except on old Solaris compilers that mangled things
incorrectly.)

src/Beman/InplaceVector/inplacevector.cpp
```diff
@@ -1,3 +1 @@
-int main() {
```
inbal2l, line 1: Is the case inconsistency with the .hpp file on the filename intentional?

src/Beman/InplaceVector/CMakeLists.txt
```diff
@@ -2,29 +2,40 @@
...
 string(TOLOWER ${CMAKE_PROJECT_NAME} CMAKE_LOWER_PROJECT_NAME)
 
+# Install the header files to the appropriate destination
 install(
         DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${CMAKE_LOWER_PROJECT_NAME}
         FILES_MATCHING
-        PATTERN "*.hpp")
+        PATTERN "*.hpp"
+)
```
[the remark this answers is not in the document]
camio: I believe the best practice is to use FILE_SET as part of the target in CMake so the exact files get installed.

src/Beman/InplaceVector/tests/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 
 # Tests
 add_executable(
-        InplaceVectorTest.test)
+        InplaceVectorTest.test InplaceVector.test.cpp)
```
camio, line 8:
```suggestion
        Beman.InplaceVectorTest.test InplaceVector.test.cpp)
```
All targets need a Beman prefix in case this is incorporated in a larger CMake project. (targets are in a global namespace).

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
+// A ContiguousContainer is a Container that stores objects in contiguous memory
+// locations.
+
+#include <algorithm>
+#include <cassert>
+#include <compare>
+#include <concepts>
+#include <iostream>
```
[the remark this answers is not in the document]
JeffGarland: good point -- iostream is a bad thing to drag in bc of static objects cout, cerr etc.  They would prevent use in an embedded context.

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
+// A ContiguousContainer is a Container that stores objects in contiguous memory
+// locations.
+
```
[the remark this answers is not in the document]
JeffGarland: This might be a policy question for Beman as well -- do we want to distinguish from the standard or do we want minimal translation from beman to std

src/Beman/InplaceVector/tests/InplaceVector.test.cpp
```diff
@@ -0,0 +1,88 @@
+#include <Beman/InplaceVector/InplaceVector.hpp>
```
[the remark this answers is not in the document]
JeffGarland: Another thread on this elsewhere.  I have to say i dislike the camel case in directory/files very much -- every press of the shift key slows things down.

Include/Beman/InplaceVector/InplaceVector.hpp
Hels15: If I understand it right, the `if constexpr (Capacity == 0) assert` should be removed from `all the assign functions`, and should be replaced with `std::bad_alloc` if the arguments passed in for the assign call have greater size than the capacity? @camio @JeffGarland

CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ project(
     DESCRIPTION "A dynamically-resizable vector with fixed capacity and embedded storage"
     LANGUAGES CXX
 )
+set(CMAKE_CXX_STANDARD 20)
```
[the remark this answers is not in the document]
Hels15: So should I remove it from the cmakelist file and explicitly provide it through the terminal?

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
+// A ContiguousContainer is a Container that stores objects in contiguous memory
+// locations.
```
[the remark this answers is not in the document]
Hels15: yes

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
...
+    If<Capacity <= std::numeric_limits<uint8_t>::max(), uint8_t,
+       If<Capacity <= std::numeric_limits<uint16_t>::max(), uint16_t,
+          If<Capacity <= std::numeric_limits<uint32_t>::max(), uint32_t,
+             uint64_t>>>;
+
+// Fixed capacity - no dynamic allocations
+template <typename T, std::size_t Capacity>
+struct inplace_vector_destruct_base {
+  using size_type = std::size_t;
+  using r_size_type = inplace_vector_size_type<Capacity>;
```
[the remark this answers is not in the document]
Hels15: Do you mean the `r_size_type` should be `internal_size_type`  as well?

Hels15:
I will need some time to resolve all these issues, thank you everyone for reviewing it.

src/Beman/InplaceVector/tests/InplaceVector.test.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+using namespace beman::InplaceVector;
```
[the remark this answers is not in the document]
Hels15: I'm not exactly sure what you mean here, where would I put "src/beman/Beman"?

Include/Beman/InplaceVector/InplaceVector.hpp
Hels15: Should all the copies use std.:ranges::copy? I guess so.

Include/Beman/InplaceVector/InplaceVector.hpp
camio: Yes, exactly.

CMakeLists.txt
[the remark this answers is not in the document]
camio: Yep

Include/Beman/InplaceVector/InplaceVector.hpp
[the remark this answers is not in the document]
camio: Yep

src/Beman/InplaceVector/tests/InplaceVector.test.cpp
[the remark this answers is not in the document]
camio: @steve-downey is using UNIX hacker jargon. "s/beman/Beman" means "everywhere you see `beman` replace it with `Beman`". The origin of this jargon is the UNIX [sed](https://en.wikipedia.org/wiki/Sed) command.

Include/Beman/InplaceVector/InplaceVector.hpp
camio: If it doesn't significantly hurt compilation performance or readability, I don't see why not.

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
...
+    std::ranges::begin(t)
+  } -> std::same_as<typename std::remove_reference_t<T>::iterator>;
+  {
+    std::ranges::end(t)
+  } -> std::same_as<typename std::remove_reference_t<T>::iterator>;
+};
+}
+template <typename T>
+constexpr bool container_compatible_range =
+    detail::container_compatible_range_impl<T>;
```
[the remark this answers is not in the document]
Hels15: I included a comment describing that this is exposition-only. I checked out what the standard has to provide but for me what I did made more sense. I do agree that this decision of mine might be a bit biased. A range must have begin() and end() and its type must be the same as the underlying iterator, that is it. If what I have is not satisfactory - I can include what the standard has. @camio

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
...
+  }
+#ifdef __cpp_lib_containers_ranges
+  template <typename R>
+    requires container_compatible_range<R>
+  constexpr inplace_vector(std::from_range_t, R &&rg) {
+    for (auto &&value : rg) {
+      emplace_back(std::forward<decltype(value)>(value));
+    }
+  }
+#else
```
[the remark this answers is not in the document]
Hels15: I'm not sure what you are suggesting here, and GitHub doesn't seem to show much.

Include/Beman/InplaceVector/InplaceVector.hpp
```diff
@@ -0,0 +1,750 @@
...
+  // [containers.sequences.inplace.vector.cons], construct/copy/destroy
+  constexpr inplace_vector() noexcept = default;
+  constexpr inplace_vector(const inplace_vector &) = default;
+  constexpr inplace_vector(inplace_vector &&) = default;
+  constexpr inplace_vector &operator=(const inplace_vector &) = default;
+  constexpr inplace_vector &operator=(inplace_vector &&) = default;
+
+  constexpr explicit inplace_vector(const size_type size) : base(size) {
+    if (size > 0) {
+      base::uninitialized_value_construct(this->begin(), this->end());
```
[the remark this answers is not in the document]
Hels15: `std:fill` will deal with already initialised(constructed) memory I think, if you track down what base(size) does you will end up with:
```cpp
  constexpr inplace_vector_destruct_base(const size_type size)
      : elems(), size_(size) {
 ```
 Here elems() does nothing because the buffer is already initialised to zero by `{}`
 ```cpp
   alignas(T) unsigned char elems[Capacity * sizeof(T)] = {};
 ```
  It basically does zero-initialisation(hence the `{}`).

File: Include/Beman/InplaceVector/InplaceVector.hpp [hunk 19]
```diff
@@ -0,0 +1,750 @@
...
+
+  // [containers.sequences.inplace.vector.cons], construct/copy/destroy
+  constexpr inplace_vector() noexcept = default;
+  constexpr inplace_vector(const inplace_vector &) = default;
+  constexpr inplace_vector(inplace_vector &&) = default;
+  constexpr inplace_vector &operator=(const inplace_vector &) = default;
+  constexpr inplace_vector &operator=(inplace_vector &&) = default;
+
+  constexpr explicit inplace_vector(const size_type size) : base(size) {
+    if (size > 0) {
```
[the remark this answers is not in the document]
Hels15: yep I got rid of them

Hels15:
@steve-downey @camio  @JeffGarland @inbal2l, Please provide me a new review/resolve them if I fixed the problem in the latest commit.

steve-downey:
Will look

Include/Beman/InplaceVector/inplace_vector.hpp
```diff
@@ -27,22 +23,24 @@ concept container_compatible_range_impl = requires(T &&t) {
...
+} // namespace detail
 template <typename T>
 constexpr bool container_compatible_range =
     detail::container_compatible_range_impl<T>;
 
 template <bool Condition, typename TrueType, typename FalseType>
 using If = typename std::conditional<Condition, TrueType, FalseType>::type;
 
+// The internal size type can be smaller than std::size_t when capacity allows
+// for it.
```
JeffGarland, line 35: clever

src/Beman/InplaceVector/tests/InplaceVector.test.cpp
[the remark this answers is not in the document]
JeffGarland: And to be pedantic it should have been ```s/beman/Beman/g``` ;)

Include/Beman/InplaceVector/inplace_vector.hpp
```diff
@@ -191,13 +192,13 @@ struct inplace_vector_base : public inplace_vector_destruct_base<T, Capacity> {
   template <typename Iter>
   static constexpr void uninitialized_copy(Iter first, Iter last,
                                            iterator dest) noexcept {
```
JeffGarland, line 200: I think we need conditional noexcept here

Include/Beman/InplaceVector/inplace_vector.hpp
```diff
@@ -120,16 +120,18 @@ struct inplace_vector_base : public inplace_vector_destruct_base<T, Capacity> {
     }
     // other.size is greater than size
     else {
+      // copy other vector into the current vector until it runs ouf of size
```
JeffGarland, line 122: ouf --> out

JeffGarland:
I've finished pass 2. This is looking pretty good - thx @Hels15!

Include/Beman/InplaceVector/inplace_vector.hpp
Hels15: So here I need to consider the cases where it is guaranteed that the function doesn't throw any exception. Since it does copying - I think a traditional 
```cpp
noexcept(std::is_nothrow_copy_constructible_v<typename std::iterator_traits<Iter>::value_type>)
```
should suffice.
@JeffGarland

Include/Beman/InplaceVector/inplace_vector.hpp
Hels15: fixed!

Hels15:
What is the current state of this PR now? @JeffGarland @steve-downey

JeffGarland (approved):

Include/Beman/InplaceVector/inplace_vector.hpp
```diff
@@ -0,0 +1,755 @@
...
+          If<Capacity <= std::numeric_limits<uint32_t>::max(), uint32_t,
+             uint64_t>>>;
+
+//  Base class for inplace_vector
+template <typename T, std::size_t Capacity>
+struct inplace_vector_destruct_base {
+  using size_type = std::size_t;
+  using internal_size_type = inplace_vector_internal_size_type<Capacity>;
+
+  alignas(T) unsigned char elems[Capacity * sizeof(T)] = {};
```
JeffGarland, line 49: something I notice on the second pass -- In this paper at least the recommendation is to use std::byte with alignas -- I'm guessing unsigned char is the same, but I'm not alignment expert. I don't want to hold the PR for this, but curious if the better practice now is byte.

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1413r3.pdf

Include/Beman/InplaceVector/inplace_vector.hpp
```diff
@@ -0,0 +1,755 @@
...
+  using reference = T &;
+  using iterator = T *;
+  using const_iterator = const T *;
+
+  // [containers.sequences.inplace.vector.cons], construct/copy/destroy
+  constexpr inplace_vector_base() = default;
+  inplace_vector_base(const inplace_vector_base &other) noexcept(
+      std::is_nothrow_copy_constructible_v<T>)
+      : inplace_vector_destruct_base<T, Capacity>(other.size) {
+    std::ranges::copy(other.begin(), other.end(), begin());
```
JeffGarland, line 100: Does this not work?

```std::ranges::copy(other, begin());```

File: Include/Beman/InplaceVector/inplace_vector.hpp [hunk 25]
```diff
@@ -0,0 +1,755 @@
...
+          std::ranges::copy(other.begin(), other.end(), begin());
+      // destroy unnecessary memory
+      std::destroy(new_end, end());
+    }
+    // other.size is greater than size
+    else {
+      // copy other vector into the current vector until it runs ouf of size
+      std::ranges::copy(other.begin(), other.begin() + size(), begin());
+      // copy the other half after the end of the current vector
+      std::ranges::copy(other.begin() + size(), other.end(), end());
```
JeffGarland, line 125: The 2 copies above need the 2 iterator overload even with ranges::copy -- so this is good.

Include/Beman/InplaceVector/inplace_vector.hpp
JeffGarland: That looks correct to me.

Include/Beman/InplaceVector/inplace_vector.hpp
```diff
@@ -0,0 +1,755 @@
...
+      throw std::bad_alloc();
+    }
+    return this->unchecked_emplace_back(std::move(x));
+  }; // freestanding-deleted
+  template <typename R>
+    requires container_compatible_range<T>
+  constexpr void append_range(R &&rg) {
+    auto first = rg.begin();
+    auto last = rg.end();
+    for (; first != last; ++first) {
```
JeffGarland, line 497: should we consider re-writing with range-for loop.  I think that would look something like:
```
for ( auto& current : rg ) {
    emplace_back(current);
}
```

steve-downey:
std::byte is the current model for the smallest addressable piece of
memory, so that makes sense, I suppose. It doesn't have all of the integral
things that a char has, but still allows aliasing, by fiat.

Include/Beman/InplaceVector/inplace_vector.hpp
```diff
@@ -0,0 +1,755 @@
...
+      emplace_back(*first);
+    }
+  }; // freestanding-deleted
+  constexpr void pop_back() {
```
JeffGarland, line 501: I'm also wondering about conditional noexcept here -- it's not noexcept in any form, in the standard text, but certainly could be if T is no throw destructible. Also it's not marked as freestanding deleted -- which I suppose is because no T on freestanding is allowed to throw? Worrying we might have some spec bugs here.

ben-craig: Re: freestanding
The design in the standard is to avoid originating exceptions.  If a user type throws first, then that isn't an issue.  This comes to the forefront in std::visit, as it can throw if a variant is in the valueless state.  You can only get on the valueless state when an exception is thrown though.

As for the noexceptness, I think it isn't currently noexcept because there's a precondition on the container being non-empty.  I thought we generally assumed stuff was noexcept destructible.  I think this could be marked unconditionally noexcept.

PR #5, opened 2024-09-20
Fix minor build issues

DeveloperPaul123:
After building with clang 18 on Linux, a number of build issues came up. This PR fixes those issues.


camio (approved):
This looks right to me. I'm surprised it compiled without these changes.

PR #6, opened 2024-09-20
Minor build setup changes

DeveloperPaul123:
Not sure if these changes are in line with the `Beman`  standards, but I found the CMakeLists changes to be necessary to get the project to build. Also I think it's fairly common to reserve the `out` or `build` folder in the source directory for builds, especially with build setups that use `CMakePresets.json` or `CMakeUserPresets.json`. 

If there are changes I need to make to get this accepted please let me know. Thanks!


CMakeLists.txt
```diff
@@ -4,6 +4,9 @@
 # cmake-format: on
 
 cmake_minimum_required(VERSION 3.23)
+
+set(CMAKE_CXX_STANDARD 23)
```
camio, line 8: We've been recommending that the caller of CMake specifies this variable. See https://github.com/beman-project/exemplar/blob/d44db6cfaa4f728d4e92aeda069c4b3679bb00ed/README.md?plain=1#L137

The rationale is that the caller of CMake should have the ability to set the language compilation flags for their project. Setting this variable (and the `cxx_std_23` property) takes away this control.

+CC @foonathan and @bretbrownjr since we discussed this at CppCon 2024.

DeveloperPaul123: Got it, that makes sense.

In that case does it make sense to just leave the `cxx_std_23` properly for the build targets in the project since that is required (?) for a successful build? This might make for an easier time for devs contributing to this project instead of having to remember to specify `DCMAKE_CXX_STANDARD=23` when configuring. Just my thoughts.

camio: It looks like we're between a rock and a hard place.

- If we don't set the `cxx_std_23` property, people trying to build without setting `-DCMAKE_CXX_STANDARD` will get a confusing error.
- If we set beman.inplace_vector's `cxx_std_23` property and this project is incorporated into a larger project using, say, C++17 the compilation flags of the target using beman.inplace_vector will silently be changed to use C++23.

I wonder what the best practice is in this situation. Maybe we could create a `try_compile` that checks for the feature beman.inplace_vector requires and, if it fails, emit an error suggesting `-DCMAKE_CXX_STANDARD` usage.

DeveloperPaul123: Well, consumers of the library should only be using an `INTERFACE` target right (header only)? In that case, the setting of `cxx_std_23` on the compiled targets should have no effect on consumer's build.

So one option would be to add an interface library for consumption:

```cmake
add_library(beman.inplace_vector INTERFACE)
target_include_directories(beman.inplace_vector INTERFACE 
	$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../../include>
	$<INSTALL_INTERFACE:include>
)
```

And then leave the compiled targets as is.

DeveloperPaul123: Do we want to support multiple C++ standards versions? If so, it's possible to allow users to set the value, but in the case they have not set `CMAKE_CXX_STANDARD` we can automatically detect the maximum C++ standard available with their tooling. 

I've done this in my thread pool implementation: https://github.com/DeveloperPaul123/thread-pool/blob/master/CMakeLists.txt#L12-L50

camio: > consumers of the library should only be using an `INTERFACE` target right (header only)?

Consumers may be using this library with `add_subdirectory` so they'd be using the `beman.inplace_vector` target defined in `src/beman/inplace_vector/CMakeLists.txt`. Even if they were using an `INTERFACE` library, I think it'd be surprising if the tests were built with different flags than everything else.

> Do we want to support multiple C++ standards versions?

Yes. Supporting multiple standards versions means more folks will try the library and provide feedback. 

> in the case they have not set `CMAKE_CXX_STANDARD` we can automatically detect the maximum C++ standard available with their tooling.

`CMAKE_CXX_STANDARD` has ["decay" behavior](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD) where it'll use the highest flag available that is less than what CMAKE_CXX_STANDARD is set to. So, setting CMAKE_CXX_STANDARD to the highest value (26) pretty much does this IIUC.

DeveloperPaul123: > Consumers may be using this library with add_subdirectory so they'd be using the beman.inplace_vector target defined in src/beman/inplace_vector/CMakeLists.txt. Even if they were using an INTERFACE library, I think it'd be surprising if the tests were built with different flags than everything else.

Right, what I meant is we should define an `INTERFACE` target instead of having users consume the static library. Unless there's a reason for doing this that I'm not aware of? I thought the static lib was there so that you have a build target even if `BUILD_TESTS` is off. Typically, I would imagine that consumers would want to turn off building tests if they are just interested in using `beman.inplace_vector` and I don't think we need a static lib for a template class. Please correct me if I'm wrong. 

> Yes. Supporting multiple standards versions means more folks will try the library and provide feedback.

Awesome, sounds good to me. 

> CMAKE_CXX_STANDARD has ["decay" behavior](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD) where it'll use the highest flag available that is less than what CMAKE_CXX_STANDARD is set to. So, setting CMAKE_CXX_STANDARD to the highest value (26) pretty much does this IIUC.

Ah, good to know, thanks!

camio: > we should define an INTERFACE target instead of having users consume the static library ... I don't think we need a static lib for a template class.

I agree.

My understanding is that `target_compile_features` works with `INTERFACE` and `IMPORT` targets the same way it works with library targets. It's too bad there isn't a way to disable the "add flag" behavior of this function.

DeveloperPaul123: > I agree.

Ok great. I can make that change as part of this PR.

> My understanding is that target_compile_features works with INTERFACE and IMPORT targets the same way it works with library targets. It's too bad there isn't a way to disable the "add flag" behavior of this function.

Yes that's correct, but unfortunately for interface targets, whatever flags you set are "public" on such targets. For non-interface targets, you can control the "visibility" of such flags.

DeveloperPaul123:
Ok I removed the static library and added an interface target. If you want @camio I can wait for changes to be implemented in beman-project/exemplar#30 and then integrate those in this project. Or that can just be a follow up PR. 

Note that the install rules in this project are still broken.

CMakeLists.txt
```diff
@@ -4,16 +4,47 @@
...
+string(TOLOWER ${CMAKE_PROJECT_NAME} CMAKE_LOWER_PROJECT_NAME)
+
+# Install the header files to the appropriate destination
+install(
+  DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${CMAKE_LOWER_PROJECT_NAME}
+  FILES_MATCHING
+  PATTERN
+    "${CMAKE_CURRENT_SOURCE_DIR}/include/beman/inplace_vector/inplace_vector.hpp"
+)
```
camio, line 45: Some tightening:

```suggestion
# Install the header files to the appropriate destination
install(
  DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${CMAKE_PROJECT_NAME}
  FILES_MATCHING
  PATTERN
    "${CMAKE_CURRENT_SOURCE_DIR}/include/beman/inplace_vector/inplace_vector.hpp"
)
```

I think this could be improved with the `FILE_SET` approach, but that can come in a later PR.

CMakeLists.txt
```diff
@@ -4,16 +4,47 @@
...
+  LANGUAGES CXX)
+
+add_library(beman.inplace_vector INTERFACE)
+target_include_directories(
+  beman.inplace_vector
+  INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+            $<INSTALL_INTERFACE:include>)
+
+# Add the current source directory to the Include directories
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
```
camio, line 22: I don't think this has any effect.

```suggestion
```

CMakeLists.txt
```diff
@@ -4,16 +4,47 @@
...
 
-include(CTest)
-add_subdirectory(src/beman/inplace_vector)
+# Set the target Include directories for the InplaceVector library
+target_include_directories(
+  beman.inplace_vector
+  INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
+
```
camio, line 28:
```suggestion
```

I think this is redundant to the `target_include_directories` call above.

camio:
> If you want @camio I can wait for changes to be implemented in https://github.com/beman-project/exemplar/pull/30 and then integrate those in this project. Or that can just be a follow up PR.

I'm okay with merging this one and doing the others in a follow up PR. Thanks for the contribution!

camio (approved):

PR #7, opened 2024-09-26
Update gitignore (again)

DeveloperPaul123:
I think these are the basic folders/files we can ignore for all projects. Since we don't provide CMakePresets.json, ignore the user presets will allow users to create their own preset file and use it for the project. We could even add a "starter" one in the exemplar.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-*build
-*out
+/build
+/out
 CMakeUserPresets.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 *build
+*out
+CMakeUserPresets.json


.gitignore
```diff
@@ -1 +1,3 @@
-*build
\ No newline at end of file
+*build
```
camio, line 1: The build instructions for beman.exemplar has builds going in the `build` directory so it makes sense it would be part of `.gitignore`. Is the `*` necessary?

Also, should this be `/build`?

.gitignore
```diff
@@ -1 +1,3 @@
...
+*out
```
camio, line 2: Can you explain the motivation for this entry?

.gitignore
```diff
@@ -1 +1,3 @@
...
+CMakeUserPresets.json
```
camio, line 3: Would this entry better reside in a global .gitignore file (like where `.*.swp` would go)?

camio (commented):
Thanks for the PR!

.gitignore
DeveloperPaul123: Whoops, you're right

.gitignore
DeveloperPaul123: From what I've seen, it's idiomatic at times to use `out/build` as an output directory; especially when using CMake presets. This allows you to put your installation directory inside `out` as well (for testing purposes) without polluting your system environment. 

So in some hypothetical preset:

```json
{
	"configurePresets" : [
		{
		    "name": "base",
		    "hidden": true,
		    "generator": "Ninja",
		    "binaryDir": "${sourceDir}/out/build/${presetName}",
		    "installDir": "${sourceDir}/out/install/${presetName}",
		    "cacheVariables": {
		        "BUILD_TESTING": "ON",
		        "CMAKE_CXX_STANDARD": "23",
		        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON"
		    },
		    "architecture": {
		        "strategy": "external",
		        "value": "x64"
		    },
		    "toolset": {
		        "strategy": "external",
		        "value": "x64"
		    }
		}
	]
}
```

.gitignore
DeveloperPaul123: We could, but then you rely on the developer to have that set up properly.

JeffGarland:
This looks ready to merge.

PR #8, opened 2024-09-26
Fix more build issues

DeveloperPaul123:
This fixes some further build issues I had. Tested on Windows with MSVC and Clang 18.

I also have a habit (good or bad?) of formatting documents whenever I work on them. Since there is no `.clangformat` in this repo I seem to have applied some default style. If there is a clang-format style to apply I'd be happy to add it. Otherwise, let me know if I need to undo the formatting changes.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,1 @@
+BasedOnStyle: LLVM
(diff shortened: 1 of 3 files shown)


camio (approved):
There's an open discussion about which `.clang-format` style is recommended, but the nice thing about auto formatting is that if the decision is to go with a different `.clang-format` we can update this repo relatively easily.

PR #19, opened 2024-11-17
fix size_

wusatosi:


wusatosi:
oh.... there's way more il-matching `size` access...

wusatosi:
I think I will bundle to test cases.

PR #22, opened 2024-11-18
Lower lowest supported version to C++17.

wusatosi:
This PR disables ranges and concept based on feature testing flags.
Swaps out C++ 20 specific functions with alternatives.

### DIFF
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -574,8 +574,7 @@ public:
 #ifdef __cpp_constexpr_dynamic_alloc
     auto final = std::construct_at(end(), std::forward<Args>(args)...);
 #else
-    // This is just construct_at expanded out, note this is not constexpr
-    // friendly
+    // Note placement new may not be not constexpr friendly
     auto final = ::new (end()) T(std::forward<Args>(args)...);
 #endif
     this->change_size(1);
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -574,7 +574,8 @@ public:
 #ifdef __cpp_constexpr_dynamic_alloc
     auto final = std::construct_at(end(), std::forward<Args>(args)...);
 #else
-    // Note placement new may not be not constexpr friendly
+    // Note: placement-new may not be constexpr friendly
+    // Avoiding placement-new may allow inplace_vector to be constexpr friendly
     auto final = ::new (end()) T(std::forward<Args>(args)...);
 #endif
     this->change_size(1);
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -571,10 +571,13 @@ public:
 
   template <class... Args>
   constexpr reference unchecked_emplace_back(Args &&...args) {
-    // TODO: what is the feature flag for std::construct_at ??
-    // auto final = std::construct_at(end(), std::forward<Args>(args)...);
-    // This is just construct_at expanded out.
+#ifdef __cpp_constexpr_dynamic_alloc
+    auto final = std::construct_at(end(), std::forward<Args>(args)...);
+#else
+    // This is just construct_at expanded out, note this is not constexpr
+    // friendly
     auto final = ::new (end()) T(std::forward<Args>(args)...);
+#endif
     this->change_size(1);
     return *final;
   };
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -2,18 +2,29 @@
 
 #include <algorithm>
 #include <cassert>
-#include <compare>
-#include <concepts>
 #include <exception>
 #include <iterator>
 #include <limits>
 #include <memory>
 #include <new>
-#include <ranges>
 #include <stdexcept>
 #include <type_traits>
+#include <utility>
+
+#ifdef __cpp_concepts
+#include <concepts>
+#endif
+
+#ifdef __cpp_lib_containers_ranges
+#include <ranges>
+#endif
+
+#ifdef __cpp_lib_three_way_comparison
+#include <compare>
+#endif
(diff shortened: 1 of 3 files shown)


include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -546,7 +571,10 @@ class inplace_vector : public inplace_vector_base<T, Capacity> {
 
   template <class... Args>
   constexpr reference unchecked_emplace_back(Args &&...args) {
-    auto final = std::construct_at(end(), std::forward<Args>(args)...);
+    // TODO: what is the feature flag for std::construct_at ??
+    // auto final = std::construct_at(end(), std::forward<Args>(args)...);
+    // This is just construct_at expanded out.
+    auto final = ::new (end()) T(std::forward<Args>(args)...);
```
wusatosi, line 577: How do I selectively use `construct_at` instead of placement new?

`construct_at` doesn't have a feature-test macro.

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -99,12 +111,12 @@ struct inplace_vector_base : public inplace_vector_destruct_base<T, Capacity> {
   inplace_vector_base(const inplace_vector_base &other) noexcept(
       std::is_nothrow_copy_constructible_v<T>)
       : inplace_vector_destruct_base<T, Capacity>(other.size) {
-    std::ranges::copy(other.begin(), other.end(), begin());
+    std::copy(other.begin(), other.end(), begin());
```
DeveloperPaul123, line 114: Maybe consider a calling delegate for `std::copy` that switches between `std::copy` and `std::ranges::copy` based on if C++20 is available? There may still be some advantages to using the `ranges` version when available.

wusatosi: I don't think there's any difference between `std::copy` and `std::ranges::copy` in our use case?

See std::ranges::copy in [cppreference (variant 3)](https://en.cppreference.com/w/cpp/algorithm/ranges/copy).
See std::copy in [cpprefernce (variant 1)](https://en.cppreference.com/w/cpp/algorithm/copy)

DeveloperPaul123: I don't have a strong opinion, was just a suggestion as there are (probably) small implementation differences between `std::copy` and `std::ranges::copy`; but I don't know that for sure. I think using `std::copy` everywhere is fine.

wusatosi: Alright! Let's keep it std::copy, I can add a patch if needed in the future

include/beman/inplace_vector/inplace_vector.hpp
DeveloperPaul123: Looks like it could be `__cpp_lib_constexpr_dynamic_alloc` or `__cpp_constexpr_dynamic_alloc` as `construct_at` was added in [P0784R7](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0784r7.html)

DeveloperPaul123 (commented):
Overall LGTM, nice work! If we can resolve the `construct_at` difficulties I think this is good to merge.

include/beman/inplace_vector/inplace_vector.hpp
DeveloperPaul123: No problem!

And no... at least not that I'm aware of off the top of my head. I would have to do some digging to find out.

wusatosi: Alright I guess constexpr stays at C++17 then.

wusatosi:
`construct_at` fixed, updated function:
https://github.com/beman-project/inplace_vector/blob/1eb14d798e6903ae756356b3296a4cae177d9a06/include/beman/inplace_vector/inplace_vector.hpp#L573-L583

@DeveloperPaul123 

Let me know if you want me to merge

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -259,15 +272,22 @@ class inplace_vector : public inplace_vector_base<T, Capacity> {
       base::uninitialized_copy(first, last, this->begin());
     }
   }
-#ifdef __cpp_lib_containers_ranges
+#else
+  template <class Itr> constexpr inplace_vector(Itr first, Itr last) : base() {
```
DeveloperPaul123, line 276: Do you think it's worth constraining `Itr` here using old school SFINAE?

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -546,7 +571,12 @@ class inplace_vector : public inplace_vector_base<T, Capacity> {
 
   template <class... Args>
   constexpr reference unchecked_emplace_back(Args &&...args) {
+#ifdef __cpp_constexpr_dynamic_alloc
     auto final = std::construct_at(end(), std::forward<Args>(args)...);
+#else
+    // Note placement new may not be not constexpr friendly
```
DeveloperPaul123, line 577:
```suggestion
    // Note: placement new may not be non-constexpr friendly
```

I think this is what was intended?

DeveloperPaul123 (approved):
Just a couple comments, but otherwise looks good to me. I tested it locally on Linux with GCC 12 and Clang 18

include/beman/inplace_vector/inplace_vector.hpp
wusatosi: That's a good question... 
I think we should just leave to SFINAE first....
Pre 17 template constraints are a pain...

wusatosi: Egh! I do hate writing enable_if and template magics.

I think we should leave this as is, we can come back and write constraints as cherry-on-top.

There's other things (tests and constexpr) I want to focus on.

include/beman/inplace_vector/inplace_vector.hpp
wusatosi: Ops typo here, placement new is not constexpr friendly. Thanks for double checking 😅

PR #26, opened 2024-11-20
Add fibonacci sequence generation as example

wusatosi:
Adds an example that shows generating a fibonacci sequence as a example.

Packed a very partial `size` fix to let the example compile.


examples/fibonacci.cpp
```diff
@@ -0,0 +1,49 @@
...
+ * Generates fibonacci sequence using inplace_vector.
+ * See: https://en.wikipedia.org/wiki/Fibonacci_sequence
+ */
+template <int Capacity> inplace_vector<int, Capacity> fibonacci_to(int num) {
+  assert(num < Capacity);
+
+  inplace_vector<int, Capacity> vec;
+
+  for (auto i = 0; i <= num; ++i) {
+    static std::array<int, 2> first_two = {0, 1};
```
DeveloperPaul123, line 19: Maybe move this static variable outside the loop? I don't see a reason why it needs to be declared inside the loop. It could also be made `constexpr`.

DeveloperPaul123 (commented):
Overall looks good, just some minor comments. 

Regarding the examples in general, are we mandating that they build with all the supported C++ standards that `inplace_vector` supports?

examples/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+set(ALL_EXAMPLES fibonacci)
+
+message("Examples to be built: ${ALL_EXAMPLES}")
+
+foreach(example ${ALL_EXAMPLES})
+    add_executable(beman.inplace_vector.examples.${example})
```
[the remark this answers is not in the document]
wusatosi: This is pulled from exemplar.

examples/fibonacci.cpp
wusatosi: Done.

wusatosi:
> Overall looks good, just some minor comments.
> 
> Regarding the examples in general, are we mandating that they build with all the supported C++ standards that `inplace_vector` supports?

Thanks for the review, we will eventually add more `constexpr` examples, e.g. fib with cache for i < 100 computed at compile time or smt. We will need to selectively enable it as it would not be supported on C++ 17 (unless we figure out a way to support constexpr on 17).

Example of this being done in exemplar: https://github.com/bemanproject/exemplar/pull/56

DeveloperPaul123:
> > Overall looks good, just some minor comments.
> > Regarding the examples in general, are we mandating that they build with all the supported C++ standards that `inplace_vector` supports?
> 
> Thanks for the review, we will eventually add more `constexpr` examples, e.g. fib with cache for i < 100 computed at compile time or smt. We will need to selectively enable it as it would not be supported on C++ 17 (unless we figure out a way to support constexpr on 17).
> 
> Example of this being done in exemplar: [bemanproject/exemplar#56](https://github.com/bemanproject/exemplar/pull/56)

Ok sounds good, thanks for the link. I have missed the last few sync calls so I'm a bit behind. I only asked because it would be nice to use `std::print()` instead of iostreams in the example(s). But that's just being nitpicky IMO. I don't have an issue landing this PR as is.

DeveloperPaul123 (approved):

PR #46, opened 2024-12-06
Review implementation

wusatosi:
# DO NOT MERGE

This is intend to review the reference implementation.

Please use GitHub's PR review feature to review the reference implementation.

I am busy with finals, I will jump in when I have time.

Link to paper: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p0843r14.html

Currently in working as section 23. https://eel.is/c++draft/inplace.vector

Would recommend starting at: [chapter](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p0843r14.html#containerssequencesinplacevector-Class-template-inplace_vector)

## Goals:
- Check if function signatures are consistent with Paper
- Check if function attributes are appropriate
- Check if implementation is valid (e.g. logical error)
- Check if struct properties are consistent with Paper
- Propose tests (you are allowed to add tests to review branch, I would not recommend updating the implementation itself).

## Non-goals:
- Large scale refactor
- Code style recommendation
- Adopting P3074 (unless comments made to how to leave room for P3074)

## NOTE
All edits are on an orphan branch, so I would suggest we minimize the edits to the implementation.

cc @jbab @DeveloperPaul123

### DIFF
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -1,0 +1,1034 @@
+#pragma once
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2023 Gonzalo Brito Gadeschi. All rights
+reserved.
+ * SPDX-License-Identifier: MIT
+ * SPDX-License-Identifier: Apache 2.0 with LLVM exception
+ * That is, this project is Dual Licensed.
+ *
+ * MIT LICENSE:
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF Precondition, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ * APACHE 2.0 with LLVM Exception:
+ ==============================================================================
+The LLVM Project is under the Apache License v2.0 with LLVM Exceptions:
+==============================================================================
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+    1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+    2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+    3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
+      cross-claim or counterclaim in a lawsuit) alleging that the Work
+      or a Contribution incorporated within the Work constitutes direct
+      or contributory patent infringement, then any patent licenses
+      granted to You under this License for that Work shall terminate
+      as of the date such litigation is filed.
+
+    4. Redistribution. You may reproduce and distribute copies of the
+      Work or Derivative Works thereof in any medium, with or without
+      modifications, and in Source or Object form, provided that You
+      meet the following conditions:
+
+      (a) You must give any other recipients of the Work or
+          Derivative Works a copy of this License; and
+
+      (b) You must cause any modified files to carry prominent notices
+          stating that You changed the files; and
+
+      (c) You must retain, in the Source form of any Derivative Works
+          that You distribute, all copyright, patent, trademark, and
+          attribution notices from the Source form of the Work,
+          excluding those notices that do not pertain to any part of
+          the Derivative Works; and
+
+      (d) If the Work includes a "NOTICE" text file as part of its
+          distribution, then any Derivative Works that You distribute must
+          include a readable copy of the attribution notices contained
+          within such NOTICE file, excluding those notices that do not
+          pertain to any part of the Derivative Works, in at least one
+          of the following places: within a NOTICE text file distributed
+          as part of the Derivative Works; within the Source form or
+          documentation, if provided along with the Derivative Works; or,
+          within a display generated by the Derivative Works, if and
+          wherever such third-party notices normally appear. The contents
+          of the NOTICE file are for informational purposes only and
+          do not modify the License. You may add Your own attribution
+          notices within Derivative Works that You distribute, alongside
+          or as an addendum to the NOTICE text from the Work, provided
+          that such additional attribution notices cannot be construed
+          as modifying the License.
+
+      You may add Your own copyright statement to Your modifications and
+      may provide additional or different license terms and conditions
+      for use, reproduction, or distribution of Your modifications, or
+      for any such Derivative Works as a whole, provided Your use,
+      reproduction, and distribution of the Work otherwise complies with
+      the conditions stated in this License.
+
+    5. Submission of Contributions. Unless You explicitly state otherwise,
+      any Contribution intentionally submitted for inclusion in the Work
+      by You to the Licensor shall be under the terms and conditions of
+      this License, without any additional terms or conditions.
+      Notwithstanding the above, nothing herein shall supersede or modify
+      the terms of any separate license agreement you may have executed
+      with Licensor regarding such Contributions.
+
+    6. Trademarks. This License does not grant permission to use the trade
+      names, trademarks, service marks, or product names of the Licensor,
+      except as required for reasonable and customary use in describing the
+      origin of the Work and reproducing the content of the NOTICE file.
+
+    7. Disclaimer of Warranty. Unless required by applicable law or
+      agreed to in writing, Licensor provides the Work (and each
+      Contributor provides its Contributions) on an "AS IS" BASIS,
+      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+      implied, including, without limitation, any warranties or conditions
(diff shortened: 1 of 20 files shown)


wusatosi:
Check list from paper:

- [ ] An inplace_vector meets all of the requirements of a container ([container.reqmts]), of a reversible container ([container.rev.reqmts]), of a contiguous container, and of a sequence container, including most of the optional sequence container requirements ([sequence.reqmts]). The exceptions are the push_front, prepend_range, pop_front, and emplace_front member functions, which are not provided. Descriptions are provided here only for operations on inplace_vector that are not described in one of these tables or for operations where there is additional semantic information.
- [ ] For any N, inplace_vector<T, N>::iterator and inplace_vector<T, N>::const_iterator meet the constexpr iterator requirements.
- [ ] For any N > 0, if is_trivial_v<T> is false, then no inplace_vector<T, N> member functions are usable in constant expressions.

Any member function of inplace_vector<T, N> that would cause the size to exceed N throws an exception of type bad_alloc.
- [ ] Let IV denote a specialization of inplace_vector<T, N>. If N is zero, then IV is both trivial and empty. Otherwise:
- [ ] Let IV denote a specialization of inplace_vector<T, N>. If N is zero, then IV is both trivial and empty. Otherwise:
- [ ] If is_trivially_copy_constructible_v<T> is true, then IV has a trivial copy constructor.
- [ ] If is_trivially_move_constructible_v<T> is true, then IV has a trivial move constructor.

If is_trivially_destructible_v<T> is true, then:
- [ ] IV has a trivial destructor,
- [ ] if is_trivially_copy_constructible_v<T> && is_trivially_copy_assignable_v<T> is true, then IV has a trivial copy assignment operator,
- [ ] if is_trivially_move_constructible_v<T> && is_trivially_move_assignable_v<T> is true, then IV has a trivial move assignment operator.

wusatosi:
Class definition:
```cpp
template <class T, size_t N>
class inplace_vector {
public:
// types:
using value_type             = T;
using pointer                = T*;
using const_pointer          = const T*;
using reference              = value_type&;
using const_reference        = const value_type&;
using size_type              = size_t;                 
using difference_type        = ptrdiff_t;             
using iterator               = implementation-defined; // see [container.requirements]
using const_iterator         = implementation-defined; // see [container.requirements]
using reverse_iterator       = std::reverse_iterator<iterator>;
using const_reverse_iterator = std::reverse_iterator<const_iterator>;

// [containers.sequences.inplace.vector.cons], construct/copy/destroy
constexpr inplace_vector() noexcept;
constexpr explicit inplace_vector(size_type n); // freestanding-deleted
constexpr inplace_vector(size_type n, const T& value); // freestanding-deleted
template <class InputIterator>
  constexpr inplace_vector(InputIterator first, InputIterator last); // freestanding-deleted
template <container-compatible-range<T> R>
  constexpr inplace_vector(from_range_t, R&& rg); // freestanding-deleted
constexpr inplace_vector(const inplace_vector&);
constexpr inplace_vector(inplace_vector&&)
  noexcept(N == 0 || is_nothrow_move_constructible_v<T>);
constexpr inplace_vector(initializer_list<T> il); // freestanding-deleted
constexpr ~inplace_vector();
constexpr inplace_vector& operator=(const inplace_vector& other);
constexpr inplace_vector& operator=(inplace_vector&& other)
  noexcept(N == 0 || (is_nothrow_move_assignable_v<T> && is_nothrow_move_constructible_v<T>));
constexpr inplace_vector& operator=(initializer_list<T>); // freestanding-deleted
template <class InputIterator>
  constexpr void assign(InputIterator first, InputIterator last); // freestanding-deleted
template<container-compatible-range<T> R>
  constexpr void assign_range(R&& rg); // freestanding-deleted
constexpr void assign(size_type n, const T& u); // freestanding-deleted
constexpr void assign(initializer_list<T> il); // freestanding-deleted

// iterators
constexpr iterator               begin()         noexcept;
constexpr const_iterator         begin()   const noexcept;
constexpr iterator               end()           noexcept;
constexpr const_iterator         end()     const noexcept;
constexpr reverse_iterator       rbegin()        noexcept;
constexpr const_reverse_iterator rbegin()  const noexcept;
constexpr reverse_iterator       rend()          noexcept;
constexpr const_reverse_iterator rend()    const noexcept;

constexpr const_iterator         cbegin()  const noexcept;
constexpr const_iterator         cend()    const noexcept;
constexpr const_reverse_iterator crbegin() const noexcept;
constexpr const_reverse_iterator crend()   const noexcept;

// [containers.sequences.inplace.vector.members] size/capacity
constexpr bool empty() const noexcept;
constexpr size_type size() const noexcept;
static constexpr size_type max_size() noexcept;
static constexpr size_type capacity() noexcept;
constexpr void resize(size_type sz); // freestanding-deleted
constexpr void resize(size_type sz, const T& c); // freestanding-deleted
static constexpr void reserve(size_type n); // freestanding-deleted
static constexpr void shrink_to_fit() noexcept;

// element access
constexpr reference       operator[](size_type n);
constexpr const_reference operator[](size_type n) const;
constexpr const_reference at(size_type n) const; // freestanding-deleted
constexpr reference       at(size_type n); // freestanding-deleted
constexpr reference       front();
constexpr const_reference front() const;
constexpr reference       back();
constexpr const_reference back() const;
    
// [containers.sequences.inplace.vector.data], data access
constexpr       T* data()       noexcept;
constexpr const T* data() const noexcept;

// [containers.sequences.inplace.vector.modifiers], modifiers
template <class... Args> constexpr reference emplace_back(Args&&... args); // freestanding-deleted
constexpr reference push_back(const T& x); // freestanding-deleted
constexpr reference push_back(T&& x); // freestanding-deleted
template<container-compatible-range<T> R>
  constexpr void append_range(R&& rg); // freestanding-deleted
constexpr void pop_back();

template<class... Args>
  constexpr pointer try_emplace_back(Args&&... args);
constexpr pointer try_push_back(const T& x);
constexpr pointer try_push_back(T&& x);
template<container-compatible-range<T> R>
  constexpr ranges::borrowed_iterator_t<R> try_append_range(R&& rg);
    
template<class... Args>
  constexpr reference unchecked_emplace_back(Args&&... args);
constexpr reference unchecked_push_back(const T& x);
constexpr reference unchecked_push_back(T&& x);
    
template <class... Args>
  constexpr iterator emplace(const_iterator position, Args&&... args); // freestanding-deleted
constexpr iterator insert(const_iterator position, const T& x); // freestanding-deleted
constexpr iterator insert(const_iterator position, T&& x); // freestanding-deleted
constexpr iterator insert(const_iterator position, size_type n, const T& x); // freestanding-deleted
template <class InputIterator>
  constexpr iterator insert(const_iterator position, InputIterator first, InputIterator last); // freestanding-deleted
template<container-compatible-range<T> R>
  constexpr iterator insert_range(const_iterator position, R&& rg); // freestanding-deleted
constexpr iterator insert(const_iterator position, initializer_list<T> il); // freestanding-deleted
constexpr iterator erase(const_iterator position);
constexpr iterator erase(const_iterator first, const_iterator last);
constexpr void swap(inplace_vector& x)
  noexcept(N == 0 || (is_nothrow_swappable_v<T> && is_nothrow_move_constructible_v<T>));
constexpr void clear() noexcept;
    
constexpr friend bool operator==(const inplace_vector& x, const inplace_vector& y);
constexpr friend synth-three-way-result<T>
  operator<=>(const inplace_vector& x, const inplace_vector& y);
constexpr friend void swap(inplace_vector& x, inplace_vector& y)
  noexcept(N == 0 || (is_nothrow_swappable_v<T> && is_nothrow_move_constructible_v<T>))
  { x.swap(y); }
};
```

tests/beman/inplace_vector/spec.test.cpp
```diff
@@ -0,0 +1,231 @@
...
+    struct T {
+      int a = 0;
+      int b;
+
+      T(T const &) = default;
+
+      T() = delete;
+      T(T &&) = delete;
+      // TODO: This type is also trivially destructible
+      // ~T() {}
```
wusatosi, line 89: Question: Is it possible to construct a trivially copy constructable struct that is not trivially destructable ?

wusatosi: Answer: no,
> is_trivially_copy_constructible is the same as is_trivially_constructible<T, const T&>, which in turn is requires is_constructible<T, const T&> where all operations are known to be trivial. is_constructible is defined in https://eel.is/c++draft/type.traits#meta.unary.prop-9, which refers to a variable, so the destructor is included.

jbab:
```suggestion
      // trivially copy constructible implies trivially destructible
```

tests/beman/inplace_vector/spec.test.cpp
```diff
@@ -0,0 +1,231 @@
...
+// sequence container, including most of the optional sequence container
+// requirements (23.2.4). The exceptions are the push_front, prepend_range,
+// pop_front, and emplace_front member functions, which are not provided.
+// Descriptions are provided here only for operations on inplace_- vector that
+// are not described in one of these tables or for operations where there is
+// additional semantic information.
+//
+// TODO: Test container & reversible container & optional sequence container
+// req.
+
```
jbab, line 25: I will drop a suggestion for 23.2.2.2 soon.

jbab: Seems I still cannot push anything to this branch. Can anyone help to enable this?

wusatosi: Let me add you write premission to this repo

wusatosi: I have sent you an invite. Note this does not grant you write access to other beman repos, if you wish to have the privilege of general write access, please contact leads on our discourse.

tests/beman/inplace_vector/spec.test.cpp
```diff
@@ -0,0 +1,716 @@
...
+  //     is equal to the value that rv had before this assignment.
+  //   Complexity: Linear.
+  X const v{0, 1, 2, 3};
+  auto const rv = [&v]() { return v; };
+  ASSERT_EQ(Counting::num_objects, v.size());
+  X t(5);
+  ASSERT_EQ(Counting::num_objects, v.size() + t.size());
+  t = rv();
+  EXPECT_TRUE((std::is_same_v<decltype(t = rv()), X &>));
+  EXPECT_EQ(Counting::num_objects, 2 * v.size()); // t[4] has been destroyed
```
jbab, line 224: Test fails because inplace_vector's destructor is not properly implemented.

tests/beman/inplace_vector/spec.test.cpp
```diff
@@ -0,0 +1,716 @@
...
+  //   Result: void.
+  //   Effects: Destroys every element of a; any memory obtained is deallocated.
+  //   Complexity: Linear.
+  alignas(X) std::byte storage[sizeof(X)];
+  X *pa = new (static_cast<void *>(storage)) X(5);
+  X &a = *pa;
+  ASSERT_EQ(Counting::num_objects, 5);
+  a.~X();
+  EXPECT_TRUE(std::is_void_v<decltype(a.~X())>);
+  EXPECT_EQ(Counting::num_objects, 0);
```
jbab, line 243: Test fails because inplace_vector's destructor is not properly implemented.

tests/beman/inplace_vector/spec.test.cpp
```diff
@@ -0,0 +1,716 @@
...
+// - No swap() function throws an exception.
+//   --> Specified in [inplace.vector.overview]
+// - No swap() function invalidates any references, pointers, or iterators
+//   referring to the elements of the containers being swapped.
+//   --> Waived by previous paragraph in [container.reqmts]
+
+// 3 For any N, inplace_vector<T, N>::iterator and inplace_vector<T,
+// N>::const_iterator meet the constexpr iterator requirements.
+//
+// TODO: Test this.
```
wusatosi, line 516: There are variations of constexpr requirement, and this depends on to be deprecated trivial type trait. We should split this out into individual files.

Specifically:
- <= C++26 constexpr when is_trivial<T>
- >C++26 constexpr for wider T using trivial union, omit any reference to is_trivial<T>

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -0,0 +1,1034 @@
...
+
+protected:
+  using __size_type = __smallest_size_t<__N>;
+
+private:
+  using __data_t =
+      conditional_t<!is_const_v<__T>, __aligned_storage2<__T, __N>,
+                    const __aligned_storage2<remove_const_t<__T>, __N>>;
+  __data_t __data_{}; // BUGBUG: test SIMD types
+  __size_type __size_ = 0;
```
wusatosi, line 444: we could use a union based lifetime delegation here instead of byte array based type erasure for supporting trivial union based constexpr:
```
union { __T __data_[__N]; };
__size_type __size_ = 0;
```

This should cover both non-trivial base case and trivial union based constexpr implementation.

tests/beman/inplace_vector/spec.test.cpp
jbab: The mentioned requirement refers to the _iterators_ of inplace_vector, not inplace_vector itself. The constexpr requirement for the iterators is independent from T and its triviality, there is no change planned in >C++26, afaik.

include/beman/inplace_vector/inplace_vector.hpp
jbab: I think there is no point switch to a union based approach before the issues addressed in [P3074](https://wg21.link/p3074 )are fixed. As of now, we would need to add a constructor and destructor to the union (as you write above), therefore the storage (and thus the whole inplace_vector) could no longer be trivially constructible and trivially destructible.
Note that the requirement

> If is_trivially_destructible_v<T> is true, then:
> IV has a trivial destructor

holds for any T, independent of its triviality (and thus, as of now, storage type).

jbab: A general notes on the tests:

- Tests should achieve a high (ideally full) coverage of all the branches, this includes not only runtime branches, but also "implementation branches" due to, e.g., different storage types for trivial vs non-trivial types.
- Ideally, we make the tests independent from the implementation (black box) and only derive the test cases from the specification.
- The current specification tells us that there must be at least two implementation branches for trivial and non-trivial T, so we should at least test with such types
- The requirements on special member functions provide more information, for example

  >  If is_trivially_destructible_v is true, then:
  >    IV has a trivial destructor

  this means that we need to test the destructor (at least) with 1. a trivial type, 2. a non-trivial type which is trivially destructible, 3. a non-trivial type which is not trivially destructible
- Further requirements on other other special member functions make it even more complicated
- In addition there are some special requirements (and thus an implementation branch) for size 0.

All in all, we need to determine the proper combination of types and then run the full test suite on them.
Once we have that we can fix the implementation and experiment with improvements.
(We might even be able to port back the implementation to C++20, ...)

I have some WIP with this respect which I hope I can commit shortly.

tests/beman/inplace_vector/spec.test2.experimental.cpp
```diff
@@ -0,0 +1,1360 @@
...
+#include <type_traits>
+
+namespace {
+
+// We run the tests on various element types with the help of GoogleTest's
+// typed tests. Which types shall we use?
+// We know from the specification of inplace_vector that there are separate
+// implementation branches for trivial and non-trivial types (due to constexpr
+// support). Furthermore, we have the requirements that the triviality of
+// inplace_vector's SMFs depends on the triviality of the element type T:
```
wusatosi, line 16: I hope this is not a stupid question. 
What is SMF?

include/beman/inplace_vector/inplace_vector.hpp
wusatosi: > we would need to add a constructor and destructor to the union (as you write above), therefore the storage (and thus the whole inplace_vector) could no longer be trivially constructible and trivially destructible.

You are right, I missed that.

I support backporting to C++ 20, the original aim was C++20 for trivial constexpr, C++17 for basic functionality. 

We might need to also test for const T.

tests/beman/inplace_vector/spec.test2.experimental.cpp
jbab: Thanks for asking. I'm using this all the time so I didn't even notice it's not obvious.
SMF = special member function, i.e. default constructor, copy constructor, move constructor, destructor, copy assignment, move assignment.

wusatosi: Thx for the clarification 👌

tests/beman/inplace_vector/spec.test.cpp
wusatosi: Ah I commented on the wrong lines, should be:

https://github.com/bemanproject/inplace_vector/blob/06855a2d83eca8a4054730cf0737b0eb593217e3/tests/beman/inplace_vector/spec.test.cpp#L518-L521

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -0,0 +1,1034 @@
...
+namespace std {
+#if defined(__GLIBCXX__) || defined(__GLIBCPP__)
+struct from_range_t {};
+inline constexpr from_range_t from_range;
+#endif
+}; // namespace std
+*/
+
+// TODO(River): using namespace std below, we need to specify
+// beman::from_range_t
```
JeffGarland, line 302: so the actual check if gcc has this is gcc14 and in c++23 mode. Or perhaps better, the feature macro #ifdef __cpp_lib_containers_ranges  see also https://en.cppreference.com/w/cpp/ranges/from_range

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -0,0 +1,1034 @@
...
+
+// Private utilities
+namespace beman::__iv_detail {
+
+template <class = void>
+[[noreturn]]
+static constexpr void __assert_failure(char const *__file, int __line,
+                                       char const *__msg) {
+  if consteval {
+    throw __msg; // TODO: std lib implementer, do better here
```
JeffGarland, line 316: well and even more is that this will be legal in c++26 after Poland https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3068r6.html was voted in

wusatosi: 👌 thanks for the input, I am not a huge fan of this static asset infrastructure as well. 

I will likely remove them or refactor them with static assert

include/beman/inplace_vector/inplace_vector.hpp
wusatosi: Given we have intention to backport this to C++20 and earlier, would it make sense to use a custom beman::from_range here?

wusatosi:
oh no I can't change it back :(

PR #55, opened 2025-01-15
Ensure implementation is up to spec

wusatosi:
The goal of this PR is to add basic spec tests so the implementation would be minimally useable. 

It would be fine to update the implementation itself in this branch.

See spec: https://eel.is/c++draft/inplace.vector

Meant to replace #46 which I accidentally broke.

cc: @jbab


include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -939,30 +951,36 @@ struct inplace_vector : private __iv_detail::__storage::_t<__T, __N> {
...
+    requires(__N != 0 && !is_trivially_move_constructible_v<__T> &&
+             movable<__T>)
   {
     for (auto &&__e : __x)
       emplace_back(::std::move(__e));
   }
 
   constexpr inplace_vector &operator=(const inplace_vector &__x)
-    requires(__N == 0)
+    requires(__N == 0 || __iv_detail::__trivial_copy_assignment<__T>)
```
jbab, line 978: Why is it that for assignment you are using concepts (trivial_copy/move_assignment) while for the constructors you are using type traits (is_trivially_copy/move_constructible)?

wusatosi: Its simply too much to write the full requirement here in-line.

I do think there should be a better name than `__trivial_copy_assignment`.

jbab: I'd argue it is less confusing if you list the individual requirements here. The difficulty finding a good name already indicates that __trivial_copy_assignment is not a good concept. Further disadvantages are:

- You have to look up in a different place what the concept means
- You have to get your head around why there is a concept in the first place. Is there something going on with assumption? How does this relate to the requirements from the specification?

It may be easier to write than you think:

- You can leave out is_trivially_destructible, since is_trivially_copyable already implies that (https://en.cppreference.com/w/cpp/language/classes#Trivially_copyable_class). However, I would still write it to have an easily reviewable alignment with the specification.
- You do not have to write a requires clause in the negative case at all. Overload resolution will always pick the more constrained version when it fits.

wusatosi: I think I have to include the negation part, I am unable to compile without the negation.

wusatosi: Suggestion adopted in 5b955bcca2cf7fed3b501a81050b3e240ee947f4

.github/workflows/ci_tests.yml
wusatosi, line 1: werror is reporting false-positive, and its suggestion is not very helpful...

For false-positive, see: https://github.com/bemanproject/inplace_vector/actions/runs/12812919831/job/35725745723

wusatosi:
The only thing left should be [sequence.reqmts]

tests/beman/inplace_vector/constructors.test.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  {
+    T value{8194};
+    IV device(IV::capacity(), value);
+
+    IV correct;
+    for (auto i = 0ul; i < device.size(); ++i)
+      correct.push_back(value);
+
+    EXPECT_EQ(device, correct);
```
jbab, line 85:
```suggestion
    EXPECT_EQ(std::count(device.begin(), device.end(), value), device.size());
```

tests/beman/inplace_vector/constructors.test.cpp
```diff
@@ -0,0 +1,133 @@
...
+  // Effects: Constructs an inplace_vector equal to the range [first, last).
+  // Complexity: Linear in distance(first, last).
+
+  using IV = TestFixture::IV;
+
+  auto reference = this->unique();
+  IV device(reference.begin(), reference.end());
+
+  EXPECT_EQ(device, reference);
+}
```
jbab, line 101: What this fails to test is whether the constructor works with an iterator which is InputIterator and nothing else. I.e. single pass, going forward only.

jbab (commented):
Some remarks on constructor.test.cpp. I still need to go through the other files.

tests/beman/inplace_vector/constructors.test.cpp
```diff
@@ -0,0 +1,133 @@
...
+  EXPECT_EQ(mid.size(), mid_size);
+  if (std::is_same_v<T, NonTriviallyDefaultConstructible> ||
+      std::is_same_v<T, NonTrivial>) {
+
+    IV mid_correct;
+    for (auto i = 0ul; i < mid_size; ++i)
+      mid_correct.emplace_back();
+
+    EXPECT_EQ(mid, mid_correct);
+  }
```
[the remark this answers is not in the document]
JeffGarland: We have this project now set to c++23 -- I'd suggest we simplify this code using ranges algo

```
EXPECT_EQ(std::ranges::count(mid, T{}), mid.size());
```

tests/beman/inplace_vector/constructors.test.cpp
```diff
@@ -0,0 +1,133 @@
...
+  if (IV::capacity() == 0)
+    return;
+
+  {
+    IV device(beman::from_range, reference | std::ranges::views::take(1));
+    EXPECT_EQ(device, IV{reference.front()});
+  }
+
+  {
+    auto mid = std::midpoint(0ul, reference.size());
```
JeffGarland, line 127: `0ul` is preferably `0uz` in c++23 to ensure it's size_t

wusatosi: thanks for the input!

tests/beman/inplace_vector/constructors.test.cpp
wusatosi: Ah I see, is there any standard library facility to warp an existing iterator to make a single pass forward only iterator?

tests/beman/inplace_vector/constructors.test.cpp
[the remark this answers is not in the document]
wusatosi: I see, I didn't realize this is value-initialized, you are right here.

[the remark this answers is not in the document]
wusatosi: Only NonTriviallyDefaultConstructible and NonTrivial have their member variables defined.

```
// A type which is not trivially default constructible (and thus not trivial),
// and all other SMFs are trivial.
struct NonTriviallyDefaultConstructible {
  int value = 0;
  friend constexpr bool operator==(NonTriviallyDefaultConstructible x, NonTriviallyDefaultConstructible y) = default;
};
```

tests/beman/inplace_vector/constructors.test.cpp
wusatosi: Thanks, adopted!

tests/beman/inplace_vector/constructors.test.cpp
wusatosi: I ended up reverting this, we would like to eventually aim to support C++20.

tests/beman/inplace_vector/constructors.test.cpp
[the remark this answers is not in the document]
JeffGarland: > The problem here is that types other than this two does not have defined value: e.g.:
> 
> ```
> // A trivial type
> struct Trivial {
>   int value;
>   friend constexpr bool operator==(Trivial x, Trivial y) = default;
> };
> ```
> 
> The value is not defined for implicit default constructor. Equality check here is invalid and undefined.

Make it defined is easy and this should be trivial?

```
struct Trivial {
   int value = 0;
   bool operator==( const Trivial& ) const = default;
};
```
https://godbolt.org/z/63nc3hqGr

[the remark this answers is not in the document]
wusatosi: > Make it defined is easy and this should be trivial?

This is not, see: https://godbolt.org/z/7GEYjEoaP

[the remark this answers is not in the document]
wusatosi: Here's how I think about this:

An array of trivial type should be allocated by just increasing the stack pointer and nothing else. So the values in the struct should be undefined.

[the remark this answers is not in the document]
JeffGarland: Sure, so basically there's no testing this because anything trivial can't do any initialization and so any comparison will be UB -- no matter the technique used to compare.

tests/beman/inplace_vector/constructors.test.cpp
wusatosi: You're right here

tests/beman/inplace_vector/constructors.test.cpp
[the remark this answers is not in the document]
wusatosi: Exactly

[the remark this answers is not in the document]
jbab: @JeffGarland, @wusatosi you are right, I forgot about my own implementation of the test types, of course we need a condition here. 
What bothers me a bit about the current condition

    if (std::is_same_v<T, NonTriviallyDefaultConstructible> ||
      std::is_same_v<T, NonTrivial>) 

is that it mentions the test types directly. We may add to them or remove some of them in the future or rename them, so this becomes brittle.
So I'm suggesting to focus on the type properties which are relevant here. This is obviously

    !std::is_trivially_default_constructible_v<T>

But there are also trivially default constructible types which can be value-initialised to a defined value (all zeros). These are the scalar types. So or condition becomes

    if (std::is_scalar_v<T> || !std::is_trivially_default_constructible_v<T>) ...

Agree?

tests/beman/inplace_vector/constructors.test.cpp
jbab: Not that I'm aware of.

wusatosi: Alright noted.

tests/beman/inplace_vector/constructors.test.cpp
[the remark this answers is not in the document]
wusatosi: This makes much more sense, thank you so much for your input. I will apply your suggestion once I have time.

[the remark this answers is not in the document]
jbab: In addition, if T is an _aggregate_ type, the expression `T{}` performs [_aggregate initialization_](https://en.cppreference.com/w/cpp/language/aggregate_initialization), which means that the member `value` will be initialized to zero. There are a few types in the test suite that qualify for this.

PR #56, opened 2025-01-17
Add missing header: cstdlib

TedLyngmo:
Also changed the build instruction from `-DCMAKE_CXX_STANDARD=20` to `-DCMAKE_CXX_STANDARD=23`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ TODO: tested platforms.
 
 ```text
 # Configure build
-$ cmake -S . -B build -DCMAKE_CXX_STANDARD=20
+$ cmake -S . -B build -DCMAKE_CXX_STANDARD=23
 -- The CXX compiler identification is GNU 11.4.0
 -- Detecting CXX compiler ABI info
 -- Detecting CXX compiler ABI info - done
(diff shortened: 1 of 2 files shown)


DeveloperPaul123 (approved):

README.md
```diff
@@ -74,7 +74,7 @@ TODO: tested platforms.
 
 ```text
 # Configure build
-$ cmake -S . -B build -DCMAKE_CXX_STANDARD=20
+$ cmake -S . -B build -DCMAKE_CXX_STANDARD=23
```
JeffGarland, line 77: What was the motivation on changing the standard level to 23?  I didn't think this implementation *required* anything in 23 -- except maybe constexpr which probably depends on 26.

TedLyngmo: @JeffGarland `consteval` is the thing

JeffGarland: ok - should that be conditional in the implementation?

TedLyngmo: No, it will not compile in pre C++23 mode so, C++23 unconditionally.

JeffGarland: Well, I'm sure with enough macro magic that wouldn't be the case.  Obviously with *degraded functionality*.  And to be clear, not trying to be difficult but just explore what's reasonable/possible.  Anyway, all good please proceed.

JeffGarland (approved):

README.md
wusatosi: We changed implementation. C++20 support in doc was for the previous implementation, I must have missed this when switching implementation. I will explore this after #55 is done.

wusatosi:
Thanks!

wusatosi (approved):
Thank you for your contribution!

PR #57, opened 2025-01-20
specify namespace std

wusatosi:
This removes use of `using namespace std`.

Waiting for #55 so we can make sure we don't break anything.

### DIFF
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -464,9 +464,9 @@ public:
   constexpr __non_trivial &operator=(__non_trivial &&) noexcept = default;
 
   constexpr ~__non_trivial()
-    requires(is_trivially_destructible_v<__T>)
+    requires(std::is_trivially_destructible_v<__T>)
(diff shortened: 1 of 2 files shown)


include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -497,8 +498,8 @@ struct inplace_vector : private __iv_detail::__storage::_t<__T, __N> {
   using difference_type = ptrdiff_t;
```
DeveloperPaul123, line 498:
```suggestion
  using difference_type = std::ptrdiff_t;
```

DeveloperPaul123, line 498:
```suggestion
  using difference_type = std::ptrdiff_t;
```

DeveloperPaul123 (approved):
Just one minor comment

include/beman/inplace_vector/inplace_vector.hpp
wusatosi: Thanks for catching that!

Issue #53, opened 2024-12-23
__smallest_size_t might be a huge pessimization on most arhitectures


cristi1990an:
While a sound idea in theory, I fear that this optimization's impact on performance far outweights any benefits it brings in size. In the restricted testing I've done, for repeated calls to push_back, using an uint8_t/uint16_t as the size resulted in more than 10 times worse performance than when using std::size_t. I feel like this is something that should be investigated a bit if the implementation is to be used in a performance-critical context.

cristi1990an:
```
#include <chrono>
#include <print>
#include "inplace_vector.hpp"

struct trivial
{
	trivial() = default;
	trivial(const trivial&) = default;
	trivial& operator=(const trivial&) = default;
};

struct non_trivial
{
	non_trivial() {}
	non_trivial(const non_trivial&) { }
	non_trivial& operator=(const non_trivial&) { return *this; };
};

static_assert(sizeof(beman::inplace_vector<trivial, 4>) == 5);
static_assert(sizeof(beman::inplace_vector<trivial, 4, false>) == 16);
static_assert(sizeof(beman::inplace_vector<non_trivial, 4>) == 5);
static_assert(sizeof(beman::inplace_vector<non_trivial, 4, false>) == 16);

template <typename T, std::size_t Size, bool SizeOptimization>
auto benchmark_loop()
{
	volatile std::size_t sum = 0; // so the below code isn't optimized away
	const auto start = std::chrono::high_resolution_clock::now();

	beman::inplace_vector<T, Size, SizeOptimization> vec;

	for (auto _ : std::views::iota(std::size_t{}, Size))
	{
		sum += vec.size();
		vec.emplace_back();
	}

	const auto stop = std::chrono::high_resolution_clock::now();
	return std::make_pair((stop - start).count(), sum);
}

template <typename T, std::size_t Size, bool SizeOptimization, std::size_t Runs>
auto benchmark()
{
	volatile std::size_t sum = 0;
	for (auto _ : std::views::iota(std::size_t{}, Runs))
	{
		sum += benchmark_loop<T, Size, SizeOptimization>().first;
	}
	return sum / Runs;
}

template <typename T, std::size_t Size, std::size_t Runs>
void display_benchmark(std::string_view type_name)
{
	std::println("Benchmark result with T = {}, Size = {} ({} runs):\n"
		"With size opt: {} Without size opt: {}\n"
		"Without size opt: {} With size opt: {}\n",
		type_name, Size, Runs,
[...]

PR #66, opened 2025-02-01
Add build option to turn off dynamic size type selection

wusatosi:
This is added because there maybe a performance penalty due to dynamic size storage that we will need to investigate, see #53 .

Todo:
- [x] Documentation 
- [x] CI Test

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,24 @@ Test project /.../inplace_vector/build
 Total Test time (real) =   0.01 sec
 ```
 
+## Build arguments
+
+### Disable dynamic size type in control block
+
+By default,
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -204,6 +204,24 @@ Test project /.../inplace_vector/build
...
 ```
 
+## Build arguments
+
+### Disable dynamic size type in control block
+
+By default,
+the type of the size variable in the control block is selected to the smallest
+it could be
+(as an upper bound can be established given we know the capacity of the vector).
```
DeveloperPaul123, line 214: Is there a reason for the way the lines are broken here?

wusatosi:
I will revisit this later.

PR #68, opened 2025-02-23
Add update regarding constexpr support

wusatosi:
As trivial unions is merged into the latest C++ standard, inplace_vector's contexpr support has been changed.
This PR documents the status of our support for trivial union.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ There have been updates to the spec after the paper getting accepted.
 Notably,
 [trivial unions (P3074)](https://github.com/cplusplus/draft/pull/7680/files#diff-3b5851f7056b7c68cb3ba2ab51cf0b75a780c6a11e5e181770067700454ace7b)
 have made the constexpr unconditional.
-This have not been implemented yet due to limited compiler support.
+constexpr has not yet been fully implemented due to limited compiler support - see [Alternative implementation using P3074 when compiler support is available](https://github.com/bemanproject/inplace_vector/issues/36).
 
 You can follow [this link](https://eel.is/c++draft/inplace.vector)
 to checkout the status of `inplace_vector` in the latest draft.
--- a/README.md
+++ b/README.md
@@ -29,11 +29,11 @@ which dynamic memory allocations are undesired.
 
 Current implementation implements all public interfaces defined in the paper.
 
-There have been minor updates to the wording after the paper is accepted,
-notably [P3247](https://wg21.link/P3247).
-Which changes the requirements for constexpr support.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,11 +29,11 @@ which dynamic memory allocations are undesired.
...
-There have been minor updates to the wording after the paper is accepted,
-notably [P3247](https://wg21.link/P3247).
-Which changes the requirements for constexpr support.
-This will likely be preceded with [P3074](https://wg21.link/P3074).
-These has not been implemented yet.
+There have been updates to the spec after the paper getting accepted.
+Notably,
+[trivial unions (P3074)](https://github.com/cplusplus/draft/pull/7680/files#diff-3b5851f7056b7c68cb3ba2ab51cf0b75a780c6a11e5e181770067700454ace7b)
+have made the constexpr unconditional.
+This have not been implemented yet due to limited compiler support.
```
JeffGarland, line 36:
```suggestion
Non-conditional no-except has not yet been implemented and has limited utility until compiler support is available. 
```
Why do we need compiler support to implement it?  If the compiler can't make it constexpr it will fall back to runtime code.  Further, there should be a feature test macro we could check if we wanted to keep the current code and change it if the compiler supports trivial unions.

wusatosi: Our current implementation is implemented with type-erased byte array, but to use trivial union, we will need to switch to a union based storage, there's three parts to this:

1. No one can benefit: As there's no confirmed compiler support.
2. Not testable: Because there's no confirmed compiler support yet, we cannot test our implementation in our current infrastructure. I do not think it's a good idea to push out an implementation that is not testable. This includes implementations that is only enabled with a build argument.
3. May break current invariant: This is not runtime vs constexpr code, trivial union actually is not designed for constexpr, it [lossens the default constructor and destructor characteristics](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3074r5.html#just-make-it-work) that make single-variant union a preferred initialization-deferred storage container. Simply switching to it may break current invariant.

wusatosi: > Non-conditional no-except has not yet been implemented

What are you referring here?

JeffGarland: > > Non-conditional no-except has not yet been implemented
> 
> What are you referring here?

ugh, sorry -- that isn't what I meant -- I've revised the suggestion.

JeffGarland (changes requested):
I see that this PR has been languishing -- I think my suggestion has been fixed

JeffGarland (approved):
Great -- good luck with the linter :)

PR #75, opened 2025-05-08
Changed pragma once to include guard, Added maybe_unused to unused params, Removed comment cruft

### LINKED ISSUE #73: cleanup implementation comments
specific details: 
- [ pragma once needs to be include guards ](https://github.com/bemanproject/inplace_vector/blob/a7397788c6344424e72ab4f73c565b18da2d16a3/include/beman/inplace_vector/inplace_vector.hpp#L1)
- [ turning off warnings isn't good practice ](https://github.com/bemanproject/inplace_vector/blob/a7397788c6344424e72ab4f73c565b18da2d16a3/include/beman/inplace_vector/inplace_vector.hpp#L2) - we shouldn't do this.  Either take out the unused parameter name in the signature or mark with maybe_unused.
- [ comment cruft should be removed ](https://github.com/bemanproject/inplace_vector/blob/a7397788c6344424e72ab4f73c565b18da2d16a3/include/beman/inplace_vector/inplace_vector.hpp#L566-L602)
- [ more comment cruft ](https://github.com/bemanproject/inplace_vector/blob/a7397788c6344424e72ab4f73c565b18da2d16a3/include/beman/inplace_vector/inplace_vector.hpp#L470-L497)

These should be mechanical changes that can be made quickly.

### LINKED ISSUE #79: warning C4068: unknown pragma 'GCC'
I'm getting lot of warning when using MSVC.

Intedai:
River (@wusatosi ):
fixes #79 
fixes #73


JeffGarland (approved):
This looks good to me.

~~Of course I'm realizing this repo basically doesn't have any testing turned on! ugh! New issue incoming...~~

I see you have to click approve and run workflows for the CI to run.

Intedai:
> This looks good to me.
> 
> ~Of course I'm realizing this repo basically doesn't have any testing turned on! ugh! New issue incoming...~
> 
> I see you have to click approve and run workflows for the CI to run.
> 
> edit: Forgot to say welcome @Intedai - thank you for your contribution!

Thanks! :smiley:

JeffGarland:
Please merge when ready.

DeveloperPaul123 (approved):
Looks good to me! Nice work 👍

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -615,12 +549,14 @@ struct inplace_vector
   }
 
 private: // Utilities
-  constexpr void assert_iterator_in_range(const_iterator it) noexcept {
+  constexpr void
+  assert_iterator_in_range([[maybe_unused]] const_iterator it) noexcept {
```
wusatosi, line 553: These functions are left over artifacts from the original reference implementation. iirc they're not used anywhere, `IV_EXPECT` is defined as a no-op currently, I think it's safe to remove all these artifacts

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -320,7 +320,8 @@ template <class T> struct zero_sized {
   using size_type = uint8_t;
   static constexpr T *storage_data() noexcept { return nullptr; }
   static constexpr size_type storage_size() noexcept { return 0; }
-  static constexpr void unsafe_set_size(size_t new_size) noexcept {
+  static constexpr void
+  unsafe_set_size([[maybe_unused]] size_t new_size) noexcept {
     IV_EXPECT(new_size == 0 &&
               "tried to change size of empty storage to non-zero value");
   }
```
wusatosi, line 327:
```suggestion
  static constexpr void
  unsafe_set_size(size_t) noexcept {}
```
Does this work?
This is assuming we remove all the assertion related artifact.

wusatosi (approved):
Thanks for the help, we can remove the artifacts from the assertion factilities in another PR.

JeffGarland:
> Thanks for the help, we can remove the artifacts from the assertion factilities in another PR.

we should make an issue for this work.  I *think*, without looking at the original, the intent of these are to be static asserts for preconditions which could be enabled. So if we agree that this is the case we should consider seeing if we can get a compiler with contracts and turn them into contract preconditions -- and maybe otherwise they just turn into asserts. 

See this paper https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3471r4.html for details.  Just to extract one thing:

As an example:

a.front()
Modify 23.2.4 [[sequence.reqmts]](https://wg21.link/sequence.reqmts) around paragraph 69 as indicated:
    a.front()
    69  Result: reference; const_reference for constant a.
    70 **Hardened preconditions: a.empty() is false.**
    71 Returns: *a.begin()
    72 Remarks: Required for basic_string, array, deque, forward_list, **inplace_vector**, list, and vector.

20162026 (changes requested):
there is some more comment cruft left at: 

https://github.com/bemanproject/inplace_vector/blob/c803c83a18a25cd6af4f2e26b77bb9d6807d4cba/include/beman/inplace_vector/inplace_vector.hpp#L505
https://github.com/bemanproject/inplace_vector/blob/c803c83a18a25cd6af4f2e26b77bb9d6807d4cba/include/beman/inplace_vector/inplace_vector.hpp#L520
https://github.com/bemanproject/inplace_vector/blob/c803c83a18a25cd6af4f2e26b77bb9d6807d4cba/include/beman/inplace_vector/inplace_vector.hpp#L543

wusatosi:
> we should make an issue for this work. I think, without looking at the original, the intent of these are to be static asserts for preconditions which could be enabled. So if we agree that this is the case we should consider seeing if we can get a compiler with contracts and turn them into contract preconditions -- and maybe otherwise they just turn into asserts.

As far as I know there hasn't been any compiler releases that support contracts. Let's hold this off till there's some compiler support as we currently do not have infrastructures that builds custom compilers for POCs.

Can you open an issue about this? For now I think we should simply remove all the artifact. We can do this in another PR or in the same PR.

JeffGarland:
> > we should make an issue for this work. I think, without looking at the original, the intent of these are to be static asserts for preconditions which could be enabled. So if we agree that this is the case we should consider seeing if we can get a compiler with contracts and turn them into contract preconditions -- and maybe otherwise they just turn into asserts.
> 
> As far as I know there hasn't been any compiler releases that support contracts. 

If you look on godbolt at least gcc has several experimental versions with contracts -- I'm not sure which one (if any) corresponds to P2900.

> Let's hold this off till there's some compiler support as we currently do not have infrastructures that builds custom compilers for POCs.

Again, there's nothing stopping someone from making those plain old assert checks -- contracts is just a fancier way to do that and have more control over the runtime results.

> Can you open an issue about this? For now I think we should simply remove all the artifact. We can do this in another PR or in the same PR.

Done: https://github.com/bemanproject/inplace_vector/issues/78

wusatosi:
> If you look on godbolt at least gcc has several experimental versions with contracts -- I'm not sure which one (if any) corresponds to P2900.

We have no infrastructure to test on experimental versions of gcc, this is on my road map but it's not there yet. I am still very much against putting things in that are not testable.

> Again, there's nothing stopping someone from making those plain old assert checks -- contracts is just a fancier way to do that and have more control over the runtime results.

We can bring back the assertion facilities if that's what you're suggesting.

Thanks for opening the issue.

Intedai:
Fixed merge conflict, sorry about the second commit accidentally removed the new line at the end :cry:

wusatosi:
> Fixed merge conflict, sorry about the second commit accidentally removed the new line at the end 😢

No you're fine, thanks for helping out!

There's a bit more cruft left, see @20162026 's comment.

Otherwise let me know when you think this PR is ready to merge.

Intedai:
@wusatosi  I removed the remaining cruft, I'm pretty sure the pr is ready to merge

PR #83, opened 2025-05-10
Use Testing Container

wusatosi:
# This is in progress

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -77,8 +77,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Print installed software
         run: |
-          clang++ --version
-          g++ --version
           cmake --version
           ninja --version
           ctest --version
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -60,42 +58,23 @@ jobs:
             compiler:
               cpp: g++
               c: gcc
+              tag: gnu-14
```
JeffGarland, line 61: do we have access to gcc-15 yet?  That would be a nice addition here.  It was released about a month ago so should be widely available on various linux's

wusatosi: Including gcc-15 would require us to build our own compiler if we want the assumption of testing container being based on ubuntu. This is because we grab all the compilers from package manager, and gcc-15 has not been released into mainstream package managers for ubuntu yet. Testing for testing container 1.0 is already a big of a pain so I want to get something landed first, let's push for gcc-15 later.

20162026: gcc-15 is available in `ubuntu/plucky` and `ubuntu/rolling⁠` but I don't know how stable those containers are

DeveloperPaul123 (approved):
Not my area of expertise, but nothing stands out to me. Overall, looks good 👍

Issue #81, opened 2025-05-10
operator=(initializer_list) is not implemented


wusatosi:
Function signature is:
```
 constexpr inplace_vector& operator=(initializer_list<T>);               // freestanding-deleted
```

This should be part of the [sequence.reqmts].

wusatosi:
> if you need help I can implement this and [#64](https://github.com/bemanproject/inplace_vector/issues/64)

That would be sweet, should be easy!

Issue #64, opened 2025-01-31
`try_append_range` is not implemented


wusatosi:
Signature:

```c++
    template<container-compatible-range<T> R>
      constexpr ranges::borrowed_iterator_t<R> try_append_range(R&& rg);
```

PR #84, opened 2025-05-10
implement missing functions

20162026:
implement:
- try_append_range #64 
- assign initializer_list #81

### DIFF
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -717,6 +717,16 @@ public:
     return *this;
   }
 
+  constexpr inplace_vector &operator=(std::initializer_list<T> il)
+    requires(std::constructible_from<
+                 T, std::ranges::range_reference_t<std::initializer_list<T>>> &&
+             std::movable<T>)
+  {
+    clear();
+    insert_range(begin(), il);
+    return *this;
+  }
+
   constexpr void
   swap(inplace_vector &x) noexcept(N == 0 ||
                                    (std::is_nothrow_swappable_v<T> &&
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -475,6 +475,18 @@ public:
     }
   }
 
+  template <details::inplace_vector::container_compatible_range<T> R>
+  constexpr std::ranges::borrowed_iterator_t<R> try_append_range(R &&rg)
+    requires(std::constructible_from<T, std::ranges::range_reference_t<R>>)
+  {
+    auto it = std::ranges::begin(rg);
+    const auto end = std::ranges::end(rg);
+    for (; size() != capacity() && it != end; ++it) {
+      try_emplace_back(std::forward<decltype(*it)>(*it));
+    }
+    return it;
+  }
+
   template <class... Args>
   constexpr iterator emplace(const_iterator position, Args &&...args)
     requires(std::constructible_from<T, Args...> && std::movable<T>)
--- a/tests/beman/inplace_vector/CMakeLists.txt
+++ b/tests/beman/inplace_vector/CMakeLists.txt
@@ -41,6 +41,7 @@ add_gtest(compare)
 add_gtest(constructors)
 add_gtest(size_n_data)
 add_gtest(erasure)
+add_gtest(modifiers)
 
 # constexpr test
 add_executable(beman.inplace_vector.tests.constexpr constexpr.test.cpp)
--- a/tests/beman/inplace_vector/modifiers.test.cpp
+++ b/tests/beman/inplace_vector/modifiers.test.cpp
@@ -1,6 +1,7 @@
-#include <gtest/gtest.h>
+#include <numeric>
 
 #include "gtest_setup.hpp"
+#include <gtest/gtest.h>
 
 namespace {
 // 23.3.14.5 Modifiers [inplace.vector.modifiers]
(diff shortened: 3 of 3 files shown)


20162026:
There were already some tests for [AssignInitializerList](https://github.com/bemanproject/inplace_vector/blob/00c10c325c86db86fb5332c5db50c6377ec2b49e/tests/beman/inplace_vector/container_requirements.test.cpp#L654) but the compiler probably substituted the il assignment with something else (il constructor?), so they passed successfully even without implementation

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -475,6 +475,18 @@ struct inplace_vector
...
   }
 
+  template <details::inplace_vector::container_compatible_range<T> R>
+  constexpr std::ranges::borrowed_iterator_t<R> try_append_range(R &&rg)
+    requires(std::constructible_from<T, std::ranges::range_reference_t<R>>)
+  {
+    auto it = std::ranges::begin(rg);
+    const auto end = std::ranges::end(rg);
+    for (; size() != capacity() && it != end; ++it) {
+      try_emplace_back(std::forward<decltype(*it)>(*it));
```
wusatosi, line 485: Is there a reason we use `try_emplace_back` here

wusatosi: The size invariant is guarded above.

tests/beman/inplace_vector/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ add_gtest(compare)
 add_gtest(constructors)
 add_gtest(size_n_data)
 add_gtest(erasure)
+add_gtest(modifiers)
```
wusatosi, line 44: I don't know why we did not caught this. I just realized this as well, this is embarrassing lol

JeffGarland: It's often difficult to 'see a negative'.  Also this implementation hasn't necessarily had a consistent group working on it...

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -717,6 +729,16 @@ struct inplace_vector
     return *this;
   }
 
+  constexpr inplace_vector &operator=(std::initializer_list<T> il)
+    requires(std::constructible_from<
+                 T, std::ranges::range_reference_t<std::initializer_list<T>>> &&
+             std::movable<T>)
+  {
+    clear();
```
JeffGarland, line 737: surely this is a pessimisation if the current is smaller than than the il -- of course we can't get the size of the initializer list because sometimes we're really dumb...

however we could perform the insert first and then only clear the elements needed -- hmm -- yeah idk this might be better. ugh

include/beman/inplace_vector/inplace_vector.hpp
20162026: yeah it can be replaced with unchecked_emplace_back, also as I understand std::forward is not necessary here?

```
unchecked_emplace_back(*it);
```

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -475,6 +475,18 @@ struct inplace_vector
...
+  template <details::inplace_vector::container_compatible_range<T> R>
+  constexpr std::ranges::borrowed_iterator_t<R> try_append_range(R &&rg)
+    requires(std::constructible_from<T, std::ranges::range_reference_t<R>>)
+  {
+    auto it = std::ranges::begin(rg);
+    const auto end = std::ranges::end(rg);
+    for (; size() != capacity() && it != end; ++it) {
+      unchecked_emplace_back(*it);
+    }
+    return it;
```
JeffGarland, line 487: I'm going to have to check the paper, but in the case that size() == capacity() at the time of the call we'll get the same result as in the case where the entire range is inserted -- so as a client I can't tell what happened? That doesn't seem good...

JeffGarland: Ok on cppreference it says:

An iterator pointing to the first element of rg that was not inserted into *this, or [ranges::end](http://en.cppreference.com/w/cpp/ranges/end)(rg) if no such element exists.

So in the case where the vector is full it should be begin(rg) which is correct here -- could we add a testcase?

include/beman/inplace_vector/inplace_vector.hpp
20162026: I dont think it matters much as for trivial type `clear` is just a `setsize` and for nontrivial it is `setsize+destructor call` (no memory allocation or other potential overhead).

include/beman/inplace_vector/inplace_vector.hpp
20162026: this one ?https://github.com/bemanproject/inplace_vector/blob/7e7696d06578f978e8007fafa961aead05eb30dc/tests/beman/inplace_vector/modifiers.test.cpp#L598

JeffGarland: ok sorry I missed it -- all the indirection in these test cases (not your fault) makes my brain hurt

include/beman/inplace_vector/inplace_vector.hpp
JeffGarland: well currently there's no optimization for trivial types -- which is a whole thing -- it actually destroys all the elements.  I think this would be better written as 

```
  //while il has elements --> assign to corresponding position (no destructor called)
  // if the iterator position < size() call erase on the rest of the elements
```

JeffGarland (approved):
I'm approving even though I have one outstanding comment.  This is better than it was and I'm sure there will be more passes on this code for further possible optimizations.

include/beman/inplace_vector/inplace_vector.hpp
20162026: yeah I will fix that, it definatly should have been assign_range...

also `constexpr iterator insert(const_iterator position, InputIterator first, InputIterator last)` implementation might not be correct, but I will investigate it tomorrow.

wusatosi: > edit: no it just erases, which doesn't destruct

What do you mean here

wusatosi (approved):
thank you for your contribution

Issue #72, opened 2025-05-06
compilation fails with exceptions disabled


20162026:
library fails to compile when using clang with `-fno-exceptions` .

should we replace `throw` with conditional macro that calls abord/terminate instead? or are there any better ways to deal with this issue in modern c++?

---
```
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:542:7: error: cannot use 'throw' with exceptions disabled
  542 |       throw std::bad_alloc();
      |       ^
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:674:7: error: cannot use 'throw' with exceptions disabled
  674 |       throw std::bad_alloc();
      |       ^
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:718:9: error: cannot use 'throw' with exceptions disabled
  718 |         throw std::bad_alloc();
      |         ^
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:722:9: error: cannot use 'throw' with exceptions disabled
  722 |         throw std::bad_alloc();
      |         ^
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:749:9: error: cannot use 'throw' with exceptions disabled
  749 |         throw std::bad_alloc{};
      |         ^
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:872:7: error: cannot use 'throw' with exceptions disabled
  872 |       throw std::bad_alloc{};
      |       ^
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:886:7: error: cannot use 'throw' with exceptions disabled
  886 |       throw std::bad_alloc{};
      |       ^
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:898:7: error: cannot use 'throw' with exceptions disabled
  898 |       throw std::out_of_range("inplace_vector::at");
      |       ^
/tmp/impl/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:903:7: error: cannot use 'throw' with exceptions disabled
  903 |       throw std::out_of_range("inplace_vector::at");
```

wusatosi:
I am not familiar with the no exception builds, anyone in have good @bemanproject/core-reviewers have suggestions?

neatudarius:
[...]

PR #88, opened 2025-05-11
noexceptions

20162026:
add support for builds with exceptions disabled #72


20162026:
- We can add the `BEMAN_INPLACE_VECTOR_NO_EXCEPTIONS` CMake option and generate `config.hpp`, but I would like to keep `defined(__cpp_exceptions)` as it's the industry-standard way to handle noexception build.

- current test implementation is ugly, but I dont know how to improve it while keeping gtest

wusatosi:
Thanks for the contribution as always, I will review tomorrow.

tests/beman/inplace_vector/noexceptions.test.cpp
```diff
@@ -0,0 +1,68 @@
+#include <array>
+
+bool abort_called = false;
+#define BEMAN_IV_THROW(x) abort_called = true;
```
wusatosi, line 4: Is [death test](https://github.com/google/googletest/blob/main/docs/advanced.md#death-tests) available to us?

wusatosi (changes requested):
See comments

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -10,19 +10,27 @@
...
 
+#if !defined(__cpp_exceptions) || __cpp_exceptions < 199711L
+#ifndef BEMAN_IV_THROW
+#include <cstdlib> // for abort
+#define BEMAN_IV_THROW(x) abort()
+#endif
+#else
+#include <stdexcept> // for length_error
+#define BEMAN_IV_THROW(x) throw x
+#endif
```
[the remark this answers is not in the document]
20162026: the idea behind `#ifndef BEMAN_IV_THROW` was to allow the user to override default abort behaviour, but I guess it will be better to check it at the top:

```suggestion
#ifndef BEMAN_IV_THROW_OR_ABORT
#if !defined(__cpp_exceptions) || __cpp_exceptions < 199711L
#include <cstdlib> // for abort
#define BEMAN_IV_THROW_OR_ABORT(x) abort()
#else
#include <stdexcept> // for length_error
#define BEMAN_IV_THROW_OR_ABORT(x) throw x
#endif
#endif
```

wusatosi:
Hey @20162026 , we will bring up some feedback against CPP.NO_FLAG_FORKING at the weekly sync the coming Monday, I think you have some valuable experience interacting and developing with this rule, would you mind writing a short feedback here so I can forward it to our meeting agenda?

include/beman/inplace_vector/inplace_vector.hpp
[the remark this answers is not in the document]
wusatosi: Ah I see what you're trying to do now!

tests/beman/inplace_vector/noexceptions.test.cpp
20162026: Ok, I somehow missed this... I will reimplement tests using death testing.

Although I don't know if there is much sense to run death tests for all throwing functions with every type from `IVAllTypes`

20162026:
@wusatosi some arguments I could think of right now, mostly regarding this and header only libs in general:

1) I don't think that ODR violation prevention measures should be enforced for header only libraries, as it is up to the user how the compilation units that are using the library should behave.
2) For header only (especially single header) libraries, it is inconvenient to have cmake generated config files, as the end user often just wants to drag and drop the headers or use a different build system, and would be forced to manually generate the config file
3) This is more of a unique case and can most likely be avoided by splitting the freestanding impl into a different namespace. However, I have a project where I would like to use both `HOSTED` and `FREESTANDING-DELETED` versions of `inplace_vector`. If we used a preprocessor define to distinguish between them and enforce it at top level, it would require multiple instances of the same library.
4) From personal observation, it is common practice for all of the standard libraries to `FLAG FORK`

wusatosi:
Thanks for the comments, will forward.

wusatosi:
oh also, how do we refer to you? : )

tests/beman/inplace_vector/noexceptions.test.cpp
```diff
@@ -0,0 +1,111 @@
+#include <array>
+
+#include "gtest_setup.hpp"
+
+namespace {
+
+template <typename Param> class NoExceptions : public IVBasicTest<Param> {};
+TYPED_TEST_SUITE(NoExceptions, IVAllTypes);
```
20162026, line 8: I guess testing all of the `IVAllTypes`  is excessive. it takes >100s for some of the CI jobs to complete 💀

wusatosi: Oh no, 100s is definitely excessive for this test. Yeah all types trashes compile time.

All types is used because our implementation switches it's underlying storage between trivial and non trivial type, and there's complications between const and non const. I think we can relax for this test honestly.

20162026:
@wusatosi my name is Edgar if that's what you are asking.

Also, PR should be more or less ready, besides the NO_FLAG_FORKING compliance

tests/beman/inplace_vector/noexceptions.test.cpp
20162026: after switching to only 4 types it is <10s, is that acceptable? or should we cut it down to a single type?

wusatosi: This works! Thanks

wusatosi:
> We can add the BEMAN_INPLACE_VECTOR_NO_EXCEPTIONS CMake option and generate config.hpp, but I would like to keep defined(__cpp_exceptions) as it's the industry-standard way to handle noexception build.

Do you think the CMake option that has its default value based on if `__cpp_exceptions` is working would be acceptable?

I can set this up directly on this PR, would you mind that?

20162026:
> Do you think the CMake option that has its default value based on if `__cpp_exceptions` is working would be acceptable?

@wusatosi  do you mean checking it in the `config.hpp` or in CMake itself using something like `check_cxx_source_compiles` ? I have pushed an example of how I would imagine it.

> I can set this up directly on this PR, would you mind that?

Feel free to push any changes you see fit. 
PS. there's no need to ask, I’d just uncheck the “Allow edits by maintainers” flag on GitHub otherwise

include/beman/inplace_vector/inplace_vector.hpp
```diff
@@ -10,19 +11,28 @@
...
 #include <ranges>
-#include <stdexcept>   // for length_error
 #include <type_traits> // for aligned_storage and all meta-functions
 
 // Artifact from previous implementation, can be used as hints for optimizer
 #define IV_EXPECT(EXPR)
 
+#ifndef BEMAN_IV_THROW_OR_ABORT
+#if !defined(__cpp_exceptions) || __cpp_exceptions < 199711L ||                \
+    BEMAN_INPLACE_VECTOR_NO_EXCEPTIONS()
```
wusatosi, line 27: > wusatosi do you mean checking it in the config.hpp or in CMake itself using something like check_cxx_source_compiles ? I have pushed an example of how I would imagine it.

Yes I am referring to using `check_cxx_source_compiles` with something like:

```c++
#if !defined(__cpp_exceptions) || __cpp_exceptions < 199711L
#error "No exception Support"
#endif

void main() { throw 0; }
```

And if this does not compile, we default `BEMAN_INPLACE_VECTOR_NO_EXCEPTIONS` to ON with a message.

We then update these lines to:
```suggestion
#if BEMAN_INPLACE_VECTOR_NO_EXCEPTIONS()
```

wusatosi:
> Feel free to push any changes you see fit.
> PS. there's no need to ask, I’d just uncheck the Allow edits by maintainers flag on GitHub otherwise

Noted, I usually find it rude for people to push directly on my branch, so I always ask HAHA.

include/beman/inplace_vector/inplace_vector.hpp
20162026: imho checking it inside of the cmake is pretty useless, as it brings to real value over just having the CMake option (whole lib must have exceptions enabled or disabled)

20162026: If we cant have `__cpp_exceptions` inside of the header, I would rather have only Cmake option, while letting user to override `BEMAN_IV_THROW_OR_ABORT()` if needed

```cpp
#ifndef BEMAN_IV_THROW_OR_ABORT
#if BEMAN_INPLACE_VECTOR_NO_EXCEPTIONS()
#include <cstdlib> // for abort
#define BEMAN_IV_THROW_OR_ABORT(x) abort()
#else
#include <stdexcept> // for length_error
#define BEMAN_IV_THROW_OR_ABORT(x) throw x
#endif
#endif

```

20162026: I have pushed some changes to avoid NOFLAGFORK roadblock and updated tests accordingly.


I did try replacing `EXPECT_THROW` with `EXPECT_DEATH` when the CMake flag is on, but it extremely increases pipeline runtime...

```
baseline with custom BEMAN_IV_THROW_OR_ABORT override
13 jobs in 8m 5s

add noexcept jobs; replace EXPECT_THROW with EXPECT_DEATH
17 jobs in 21m 49s

add noexcept jobs; replace EXPECT_THROW with do{}while(0); only run noexcept tests with corresponding cmake option
17 jobs in 8m 4s
```

tests/beman/inplace_vector/gtest_setup.hpp
```diff
@@ -392,3 +392,11 @@ template <typename Param> class IVBasicTest : public ::testing::Test {
 
 template <typename Param>
 std::size_t IVBasicTest<Param>::InputIterator::num_deref;
+
+#if BEMAN_INPLACE_VECTOR_NO_EXCEPTIONS()
+#define SAFE_EXPECT_THROW(x, y)                                                \
+  do {                                                                         \
+  } while (0)
+#else
```
wusatosi, line 400: This is cool! Great job!

tests/beman/inplace_vector/CMakeLists.txt
```diff
@@ -43,6 +43,21 @@ add_gtest(size_n_data)
...
+    AND (
+        CMAKE_CXX_COMPILER_ID STREQUAL "Clang"
+        OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
+    )
+)
+    add_gtest(noexceptions)
+    target_compile_options(
+        beman.inplace_vector.tests.noexceptions
+        PRIVATE -fno-exceptions
+    )
```
wusatosi, line 58: Is this still needed?

wusatosi (approved):
Thank you so much for all the work!

I have one minor question but besides that everything looks great.

tests/beman/inplace_vector/CMakeLists.txt
20162026: Yes, it ensures that there are no throws in the compiled code. Without it, EXPECT_DEATH could still pass if an unhandled exception was thrown.


Although, I'm now not sure if we need the compiler check. It allows running the tests on compilers that don't support or have a different syntax for disabling exceptions (e.g. MSVC). But right now, CI only includes Clang and GCC, and maybe it would be better to error out instead of silently skipping tests on other compilers. On the other hand, it doesn't matter much, as the implementation should be compiler agnostic as long as there is abort()
```c++
        CMAKE_CXX_COMPILER_ID STREQUAL "Clang"
        OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
```

PR #90, opened 2025-05-22
freestanding namespace

20162026:
alternative of #77 using separate namespace for freestanding implementation

### DIFF
--- a/tests/beman/inplace_vector/freestanding.test.cpp
+++ b/tests/beman/inplace_vector/freestanding.test.cpp
@@ -1,0 +1,324 @@
+#include <array>
+#include <beman/inplace_vector/inplace_vector.hpp>
+#include <gtest/gtest.h>
+
+// constexpr void assign(InputIterator first, InputIterator last);
+template <typename, typename = std::void_t<>>
+struct has_assign_iterator : std::false_type {};
+template <typename T>
+struct has_assign_iterator<T, std::void_t<decltype(std::declval<T>().assign(
+                                  std::declval<typename T::iterator>(),
+                                  std::declval<typename T::iterator>()))>>
+    : std::true_type {};
+
+// constexpr void assign_range(R&& rg);
+template <typename T, typename R, typename = void>
+struct has_assign_range : std::false_type {};
+template <typename T, typename R>
+struct has_assign_range<T, R,
+                        std::void_t<decltype(std::declval<T &>().assign_range(
+                            std::declval<R &&>()))>> : std::true_type {};
+
+// constexpr void assign(size_type n, const T& u);
+template <typename T, typename = void>
+struct has_assign_size : std::false_type {};
+template <typename T>
+struct has_assign_size<T, std::void_t<decltype(std::declval<T>().assign(
+                              std::declval<typename T::size_type>(),
+                              std::declval<typename T::const_reference>()))>>
+    : std::true_type {};
+
+// constexpr void assign(initializer_list il);
+template <typename T, typename = void>
+struct has_assign_initializer : std::false_type {};
+template <typename T>
+struct has_assign_initializer<
+    T, std::void_t<decltype(std::declval<T>().assign(
+           std::declval<std::initializer_list<typename T::value_type>>()))>>
+    : std::true_type {};
+
+// constexpr void resize(size_type sz);
+template <typename T, typename = void> struct has_resize : std::false_type {};
+template <typename T>
+struct has_resize<T, std::void_t<decltype(std::declval<T>().resize(
+                         std::declval<typename T::size_type>()))>>
+    : std::true_type {};
+
+// constexpr void resize(size_type sz, const T& c);
+template <typename T, typename = void>
+struct has_resize_ref : std::false_type {};
+template <typename T>
+struct has_resize_ref<T, std::void_t<decltype(std::declval<T>().resize(
+                             std::declval<typename T::size_type>(),
+                             std::declval<typename T::const_reference>()))>>
+    : std::true_type {};
+
+// void reserve(size_type n);
+template <typename T, typename = void> struct has_reserve : std::false_type {};
+template <typename T>
+struct has_reserve<T, std::void_t<decltype(std::declval<T>().reserve(
+                          std::declval<typename T::size_type>()))>>
+    : std::true_type {};
+
+// constexpr reference at(size_type n);
+template <typename T, typename = void> struct has_at : std::false_type {};
+template <typename T>
+struct has_at<T, std::void_t<decltype(std::declval<T>().at(
+                     std::declval<typename T::size_type>()))>>
+    : std::true_type {};
+
+// constexpr const_reference at(size_type n) const;
+template <typename T, typename = void> struct has_const_at : std::false_type {};
+template <typename T>
+struct has_const_at<T, std::void_t<decltype(std::declval<const T>().at(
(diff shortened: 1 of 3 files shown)


wusatosi:
btw, current single-header-ness is not intentional, feel free to break the current header into multiple header files.

20162026:
> btw, current single-header-ness is not intentional, feel free to break the current header into multiple header files.

that would open up a whole new discussion on how we would want to split it :D and I would rather leave it for a different (refactoring) PR


As for this PR, I'm also tempted to add some convenience functions for the freestanding implementation (e.g., `initializer_list` constructor and from/to non-freestanding `assignment operators`). We could also add them under a CMake option that is off by default. **But first, I would like to hear some feedback on whether this approach is suitable**, or if there are any edge cases I haven't thought through...

wusatosi:
> As for this PR, I'm also tempted to add some convenience functions for the freestanding implementation (e.g., `initializer_list` constructor and from/to non-freestanding `assignment operators`). We could also add them under a CMake option that is off by default. **But first, I would like to hear some feedback on whether this approach is suitable**, or if there are any edge cases I haven't thought through...

Yeah, also there's no none-freestanding-delete insert operation...
We should definitively keep one in.

I say we should get the first minimal implementation started, we can then gradually decide which function our implementation want to keep, we definitely want to at least have "no freestanding-delete functions", "some freestanding-delete functions", "all freestanding-delete functions".

wusatosi:
The problem is if we want to do this gradually, we might might-as-well consider continuing with the approach in #77 . So I think it will be helpful for us to scratch out the approach.

20162026:
> The problem is if we want to do this gradually, we might might-as-well consider continuing with the approach in #77 . So I think it will be helpful for us to scratch out the approach.

With the previous approach, having to globally choose either a freestanding or non-freestanding implementation is a dealbreaker for me. Of course, I could just fork and do whatever I want, but I would rather avoid it, especially with such a big changes that would be hard to sync.

> I say we should get the first minimal implementation started, we can then gradually decide which function our implementation want to keep, we definitely want to at least have "no freestanding-delete functions", "some freestanding-delete functions", "all freestanding-delete functions" as I think current "no freestanding-delete functions" are non-ergonomic.

I guess it depends on what the end goal of a freestanding implementation is. For me, it's about getting compile time warnings when trying to use any potentially unsafe functions, but in theory there's more to freestanding than just that...

wusatosi:
I mean if your intention is to have an inplace vector that does not have any actively throwing functions,

I believe you need to delete any functions that [increase the size of the vector without checking in advance](https://eel.is/c++draft/inplace.vector#overview-4.sentence-1). Along with `at`.

You will still have any transitive exception like when the copy constructor throws, but I don't believe that can be completely mitigated through deleting freestanding functions.

Which I think is the current list of freestanding-deleted functions.

wusatosi:
If no throwing function is the intention, then having the standalone struct is actually a better idea. But I think if we are trying to play the line of "ah but this function has great utility, we should leave it in", we will have throwing functions otherwise the implementation will be non-compliance.

I think I will reverse my previous stance, I am for a separate struct that is in the freestanding namespace, and I am for the struct to delete all freestanding-delete functions without leaving in convenience functions.

20162026:
> I am for a separate struct that is in the freestanding namespace, and I am for the struct to delete all freestanding-delete functions without leaving in convenience functions

I'm also fine with deleting all of them. Convenience functions are definitely not mandatory.

wusatosi:
> I'm also fine with deleting all of them. Convenience functions are definitely not mandatory.

Let's do that, if someone wants non-throwing convenience function in a restricted context they can disable exception.

tests/beman/inplace_vector/freestanding.test.cpp
```diff
@@ -0,0 +1,330 @@
...
+                                  std::declval<typename T::iterator>()))>>
+    : std::true_type {};
+
+// constexpr void assign_range(R&& rg);
+template <typename T, typename R, typename = void>
+struct has_assign_range : std::false_type {};
+template <typename T, typename R>
+struct has_assign_range<T, R,
+                        std::void_t<decltype(std::declval<T &>().assign_range(
+                            std::declval<R &&>()))>> : std::true_type {};
```
wusatosi, line 20: I have never seen this before, how does this work?

20162026: SFINAE magic that checks if the function is callable and returns true, or substitutes with false if the function does not exist.
I guess we can rewrite all of the checks using concepts to be cleaner and more up to date with modern C++.

wusatosi: No you can keep it I just never seen this lol

20162026: For the sake of sanity of anyone who has to read the tests, I will still rewrite it using concepts

JeffGarland (approved):
I'm going to approve this PR as it looks like it's basically ready.  The one big gaping omission here is an update to readme regarding the freestanding option.  Of course that can be done in a separate PR so I won't hold this up for that.

I'm going to continue to point out that calling the freestanding implementation `freestanding-deleted` is backwards.  The implementation is a `freestanding implementation`, not `freestanding-deleted implementation`.  Yes, the net effect of a `freestanding-implementation` is to delete functions but the only time the standard applies 'freestanding-deleted' is in the context of a single function.

wusatosi (commented):
Thank you so much for you contribution.
This is fantastic.

Looking at the implementation, I don't think we need to test both structs.

The separation between freestanding hosted and freestanding deleted test suites would be helpful for us to make sure those extra function throws as expected, but not really useful within the context of this PR.

It will be great if you can add some docs work, but we can do that in another PR.

Let me know if you want me to just merge this @20162026 .

wusatosi:
Actually, I have a question, do the inplace_vector and freestanding::inplace_vector work with each other? e.g. on assignment, copy constructor, etc... I am more than happy to leave this behavior undefined, I just wonder if we need to guard against this at all.

20162026:
> Actually, I have a question, do the inplace_vector and freestanding::inplace_vector work with each other? e.g. on assignment, copy constructor, etc... I am more than happy to leave this behavior undefined, I just wonder if we need to guard against this at all.

technically we could even allow conversion from one to another as there should be no problem as long as T and N are the same. Or just add some tests to ensue none of the tested compilers do something unexpected, e.g:

```cpp
TYPED_TEST(Constructors, freestandingConversion) {
  using T = TestFixture::T;

  using IV = beman::inplace_vector<T, 5>;
  using FS = beman::freestanding::inplace_vector<T, 5>;
  using BASE = beman::details::inplace_vector::inplace_vector_base<T, 5>;

  static_assert(std::is_constructible_v<FS, FS>);
  static_assert(std::is_constructible_v<IV, IV>);
  static_assert(std::is_convertible_v<IV, BASE>);
  static_assert(std::is_convertible_v<FS, BASE>);

  static_assert(!std::is_convertible_v<BASE, IV>);
  static_assert(!std::is_convertible_v<BASE, FS>);
  static_assert(!std::is_constructible_v<IV, FS>);
  static_assert(!std::is_constructible_v<FS, IV>);
}
```

wusatosi:
Thanks for the doc work. I don't know if we want this, the fact that someone is mixing the two should be a code smell.

wusatosi:
> @wusatosi can we merge or should we wait for another review?

Let's ship it

Issue #54, opened 2024-12-24
undefined behavior from reinterpret_cast


changkhothuychung:
https://github.com/bemanproject/inplace_vector/blob/main/include/beman/inplace_vector/inplace_vector.hpp#L364-L367
Suppose T is a user-defined type, wouldn't this reinterpret_cast result in an undefined behavior? A solution would be to use std::launder. 

```cpp
template <class __T, size_t __N> struct __aligned_storage2 {
  alignas(__T) byte __d[sizeof(__T) * __N];
  constexpr __T *__data(size_t __i) noexcept {
    __IV_EXPECT(__i < __N);
    return reinterpret_cast<__T *>(__d) + __i;
  }
  constexpr const __T *__data(size_t __i) const noexcept {
    __IV_EXPECT(__i < __N);
    return reinterpret_cast<const __T *>(__d) + __i;
  }
};
```

wusatosi:
Sorry I am not familiar with std::launder, Can you elaborate on why this would result in undefined behavior?

jbab:
There is no undefined behaviour here, https://en.cppreference.com/w/cpp/language/reinterpret_cast#Explanation, paragraph 5.
Basically we put Ts in and get Ts out. Also, the proper alignment is ensured.

Of course, it takes more than the quoted code to prove the point. The implementation of inplace_vector has to make sure that we really only put Ts in. This is easy to review, though.

wusatosi:
@tzlaine you seem to be knowledge about object lifetime, can you help look at this issue?

PR #96, opened 2025-06-07
Remove reinterpret_cast

Halalaluyafail3:
It is not valid to use reinterpret_cast in a constant expression, and even outside of constant expressions the use is UB as described in #54. Instead of using a byte array, a union which contains an array can be used for the same purpose. This makes the name of the class wrong, but I have left it unchanged for now.

### DIFF
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -139,15 +139,17 @@ public:
   constexpr ~trivial() = default;
 };
 
-template <class T, size_t N> struct raw_byte_based_storage {
-  alignas(T) std::byte _d[sizeof(T) * N];
-  constexpr T *storage_data(size_t i) noexcept {
+template <class T, std::size_t N> union raw_byte_based_storage {
+  T _d[N];
+  constexpr raw_byte_based_storage() {}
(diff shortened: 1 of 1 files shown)


wusatosi:
Hi @Halalaluyafail3 , this implement is based on the original inplace vector paper (P0843) without implementing trivial unions (P3074) (as currently there's no compiler support).

In P0843, inplace vector is only required to be constexpr for trivial element type T.

This is designed that for non trivial types, the default constructor is not called unnecessarily and allows default constructor deleted element type.

You can read more about this here: [The storage problem](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3074r7.html#the-uninitialized-storage-problem).

Halalaluyafail3:
Ok, if that is intended it seems misleading to use constexpr in the definitions of storage_data.

wusatosi:
> Ok, if that is intended it seems misleading to use constexpr in the definitions of storage_data.

good point, it shouldn't be.

JeffGarland:
So it seems like we aren't going to merge this, but maybe the cast should have a big juicy comment on it that explains why it's there for now?

wusatosi:
See #102 .

PR #97, opened 2025-06-09
Fix gcc15 const T storage

### LINKED ISSUE #91: ref_impl.text.cpp compilation fails g++-15
This seems to be something new with g++-15 as the current code compiles and runs with g++-14.  This is on ubuntu arm64.

I haven't had time to dig into the compiler complaint here, but I'm going to assume it's reporting a real attempt to strip const from void*.

```
cmake -B build -S . -DCMAKE_CXX_COMPILER=g++-15 -DCMAKE_CXX_STANDARD=23 -DCMAKE_VERBOSE_MAKEFILE=ON


[ 11%] [32mBuilding CXX object tests/beman/inplace_vector/CMakeFiles/beman.inplace_vector.ref-test.dir/ref_impl.test.cpp.o[0m
cd /home/jeff/dev/inplace_vector/build/tests/beman/inplace_vector && /usr/bin/g++-15  -I/home/jeff/dev/inplace_vector/include -I/home/jeff/dev/inplace_vector/build/include -std=gnu++23 -MD -MT tests/beman/inplace_vector/CMakeFiles/beman.inplace_vector.ref-test.dir/ref_impl.test.cpp.o -MF CMakeFiles/beman.inplace_vector.ref-test.dir/ref_impl.test.cpp.o.d -o CMakeFiles/beman.inplace_vector.ref-test.dir/ref_impl.test.cpp.o -c /home/jeff/dev/inplace_vector/tests/beman/inplace_vector/ref_impl.test.cpp
In file included from /usr/include/c++/15/bits/stl_iterator.h:78,
                 from /usr/include/c++/15/bits/stl_algobase.h:67,
                 from /usr/include/c++/15/algorithm:62,
                 from /home/jeff/dev/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:10,
                 from /home/jeff/dev/inplace_vector/tests/beman/inplace_vector/ref_impl.test.cpp:18:
/usr/include/c++/15/bits/stl_construct.h: In instantiation of ‘constexpr _Tp* std::construct_at(_Tp*, _Args&& ...) [with _Tp = const int; _Args = {const int&}]’:
/home/jeff/dev/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:352:22:   required from ‘constexpr T& beman::inplace_vector<T, N>::unchecked_emplace_back(Args&& ...) requires  constructible_from<T, Args ...> [with Args = {const int&}; T = const int; long unsigned int N = 3]’
  352 |     std::construct_at(end(), std::forward<Args>(args)...);
      |     ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jeff/dev/inplace_vector/include/beman/inplace_vector/inplace_vector.hpp:399:34:   required from ‘constexpr T& beman::inplace_vector<T, N>::unchecked_push_back(const T&) requires  constructible_from<T, const T&> [with T = const int; long unsigned int N = 3]’
  399 |     return unchecked_emplace_back(x);
      |            ~~~~~~~~~~~~~~~~~~~~~~^~~
/home/jeff/dev/inplace_vector/tests/beman/inplace_vector/ref_impl.test.cpp:84:24:   required from here
   84 | template struct beman::inplace_vector<const int, 3>;
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/15/bits/stl_construct.h:99:21: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
   99 |       void* __loc = __location;
      |                     ^~~~~~~~~~
      |                     |
      |                     const void*
```

20162026:
fix #91 

tested with `g++-15 (Ubuntu 15-20250404-0ubuntu1) 15.0.1 20250404 (experimental) [master r15-9193-g08e803aa9be]` on `ubuntu:rolling` docker img

Im not sure if this approach is correct or why the previous impl used const container for `const T` storage

### DIFF
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -112,10 +112,7 @@ protected:
   using size_type = smallest_size_t<N>;
 
 private:
-  // If value_type is const, then const std::array of non-const elements:
-  using array_based_storage =
-      std::conditional_t<!std::is_const_v<T>, std::array<T, N>,
-                         const std::array<std::remove_const_t<T>, N>>;
+  using array_based_storage = std::array<std::remove_const_t<T>, N>;
   alignas(alignof(T)) array_based_storage storage_data_{};
   size_type storage_size_ = 0;
 
@@ -123,7 +120,7 @@ protected:
   constexpr const T *storage_data() const noexcept {
     return storage_data_.data();
   }
-  constexpr T *storage_data() noexcept { return storage_data_.data(); }
+  constexpr auto storage_data() noexcept { return storage_data_.data(); }
   constexpr size_type storage_size() const noexcept { return storage_size_; }
   constexpr void unsafe_set_size(size_t new_size) noexcept {
     IV_EXPECT(size_type(new_size) <= N && "new_size out-of-bounds [0, N]");
@@ -161,9 +158,7 @@ protected:
   using size_type = smallest_size_t<N>;
 
 private:
-  using byte_based_storage = std::conditional_t<
-      !std::is_const_v<T>, raw_byte_based_storage<T, N>,
-      const raw_byte_based_storage<std::remove_const_t<T>, N>>;
+  using byte_based_storage = raw_byte_based_storage<std::remove_const_t<T>, N>;
   byte_based_storage storage_data_{}; // BUGBUG: test SIMD types
   size_type storage_size_ = 0;
 
@@ -171,7 +166,9 @@ protected:
   constexpr const T *storage_data() const noexcept {
     return storage_data_.storage_data(0);
   }
-  constexpr T *storage_data() noexcept { return storage_data_.storage_data(0); }
+  constexpr auto storage_data() noexcept {
+    return storage_data_.storage_data(0);
+  }
   constexpr size_type storage_size() const noexcept { return storage_size_; }
   constexpr void unsafe_set_size(size_t new_size) noexcept {
     IV_EXPECT(size_type(new_size) <= N && "new_size out-of-bounds [0, N)");
@@ -323,7 +320,7 @@ public:
     requires(std::constructible_from<T, Args...>)
   {
     IV_EXPECT(size() < capacity() && "inplace_vector out-of-memory");
-    std::construct_at(end(), std::forward<Args>(args)...);
+    std::construct_at(storage_data() + size(), std::forward<Args>(args)...);
     unsafe_set_size(size() + size_type(1));
     return this->back();
   }


20162026:
Also, should `beman::inplace_vector<const int, 3> cvec{1, 2, 3};` compile? Because right now it fails both in main and with this pr

wusatosi:
> Also, should beman::inplace_vector<const int, 3> cvec{1, 2, 3}; compile? Because right now it fails both in main and with this pr

It should

20162026:
> > Also, should beman::inplace_vector<const int, 3> cvec{1, 2, 3}; compile? Because right now it fails both in main and with this pr
> 
> It should

Then I guess we'll need to review the current implementation because it pretty much prevents using most constructors or functions for const T. But this is outside the scope of this PR...

also what compiler settings did you use when you got the warning with 15.1?

JeffGarland:
> no issues. So not sure why you're having an issue?

oops I see -- you must have made a change on the branch that stops the error:

/usr/include/c++/15/bits/stl_construct.h:99:21: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
   99 |       void* __loc = __location;
      |                     ^~~~~~~~~~
      |                     |
      |                     const void*

JeffGarland:
Thinking about this, I'm not seeing how `const T` can be valid for any standard container -- how would assign to a value in the first place since it's const. With vector it actually has a static_assert to prevent this case:

```
/opt/compiler-explorer/gcc-15.1.0/include/c++/15.1.0/bits/stl_vector.h: In instantiation of 'class std::vector<const int>':
<source>:13:28:   required from here
   13 |     std::vector<const int> ci { 1, 2, 3};
      |                            ^~
/opt/compiler-explorer/gcc-15.1.0/include/c++/15.1.0/bits/stl_vector.h:470:66: error: static assertion failed: std::vector must have a non-const, non-volatile value_type
  470 |       static_assert(is_same<typename remove_cv<_Tp>::type, _Tp>::value,
  ```
https://godbolt.org/z/johqKq3YW

I don't really see wording that forces this to be the case, but I might just be missing it...

20162026:
the problem with `inplace_vector` is that I dont see any requirements in the paper that would prevent or approve using `const T`, while regular vector (if I'm not mistaken) requires `T` to be `CopyAssignable`.

I dont see any real world usecase of `inplace_vector<const T, N>` but dont think we should add a `static_assert` that is not required by the paper

JeffGarland:
hmm, right -- we had to make this just impossible to discover the rules.  You're correct that a constexpr initialization works -- but array is fixed sized with non of the dynamics of inplace_vector.  Note that const isn't an option if you try to do much of anything with array either. 

https://eel.is/c++draft/containers#array.cons-1

Also it clearly didn't start this way in c++11 https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0031r0.html

JeffGarland:
Sorry we didn't have time to address at the sync -- and I saw this after the meeting.

Issue #89, opened 2025-05-22
Missing constructor test for `inplace_vector(initializer_list)`


wusatosi:
Also default constructor for inplace_vector

wusatosi:
The constructors are tested in `tests/beman/inplace_vector/container_requirements.test.cpp`

20162026:
resize test added in #100

PR #100, opened 2025-08-07
move implementation into beman::inplace_vector namespace

### LINKED ISSUE #85: namespaces are wrong
Have a look at beman.optional here:

https://godbolt.org/z/Gc6Y9j6zf

Notice:

```
#include <beman/optional/optional.hpp>
....
beman::optional::optional<int> empty_opt{};
```
The include paths and namespaces match.  Now look at inplace_vector

https://godbolt.org/z/6f7Ea8P4q

```
#include <beman/inplace_vector/inplace_vector.hpp>
...
using namespace beman::inplace_vector;
//gives the error:
<source>:6:24: error: 'inplace_vector' is not a namespace-name
    6 | using namespace beman::inplace_vector;
```
That's because inplace_vector is only in the beman namespace https://godbolt.org/z/bd3ob1bf5

The standard sez:

[CPP.NAMESPACE] RECOMMENDATION: Headers in include/beman/<short_name>/ should export entities in the beman::<short_name> namespace.

It's only a recommendation, but we should follow it in my view for consistency.

Similarly if you look at details you optional you'll see this for details. https://github.com/bemanproject/optional/blob/069cd74ff18e1f3c82d054b2f7b50dc22a926509/include/beman/optional/detail/iterator.hpp#L12

Notice the namespace is beman::optional::detail.  For inplace vector we have beman::detail

https://github.com/bemanproject/inplace_vector/blob/b8aabf6f97f65c8b6471a1daf744d7e5ab6c9e43/include/beman/inplace_vector/inplace_vector.hpp#L33

which should be beman::inplace_vector::detail - even though the standard is silent on the matter (it probably shouldn't be).

This really should be an easy mechanical change.

20162026:
close #85 

namespace changes:
`beman::inplace_vector<..>` moved to `beman::inplace_vector::inplace_vector<..>`  
`beman::details::inplace_vector` moved to `beman::inplace_vector::details`   
`beman::freestanding::inplace_vector` to `beman::inplace_vector::freestanding`  
`beman::from_range` polyfill moved to `beman::inplace_vector::details::from_range`
`beman::erase` moved to `beman::inplace_vector::erase`

this will couse some minor inconvenience for anyone who is already using the library, but I do believe we should fix the namespace for consistency

also add missing test case for resize with init value #89

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Contributions are welcome.
 
 #include <beman/inplace_vector/inplace_vector.hpp>
 
-using namespace beman;
+using namespace beman::inplace_vector;
 
 /**
  * Generates fibonacci sequence using inplace_vector.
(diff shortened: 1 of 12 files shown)


20162026:
@wusatosi if there are no objections, I'm going to merge this tomorrow

PR #106, opened 2025-12-29
remove is_trivial type trait

20162026:
Remove deprecate is_trivial type trait from inplace_vector to fix compiler warnings.

this PR has cherrypicked changes from #97 to fix test compilation under gcc15, 
to avoid potential issues it should be merged after #97

tested with `ghcr.io/bemanproject/infra-containers-gcc:latest` GCC 15.2.0 C++26


JeffGarland:
> Current CI workflow would require major overhaul to support gcc 15 as it is using old beman container registry that has no images for 15
> 
> I guess it will (or at least should be) addressed in #103

Yes, it's not really that difficult as I understand it -- optional got converted.

PR #130, opened 2026-04-28
Fix smallest_size_t

### LINKED ISSUE #123: smallest_size_t is not correct
smallest_size_t is calculated using:
```
template <size_t N>
using smallest_size_t
= std::conditional_t<(N < std::numeric_limits<uint8_t>::max()),  uint8_t,
    std::conditional_t<(N < std::numeric_limits<uint16_t>::max()), uint16_t,
    std::conditional_t<(N < std::numeric_limits<uint32_t>::max()), uint32_t,
    std::conditional_t<(N < std::numeric_limits<uint64_t>::max()), uint64_t,
                   size_t>>>>;
```
But it should be `<=` instead of `<`.
I noticed this because my test failed:
```
inplace_vector<uint8_t, 255> vector_small;
ASSERT_EQ(sizeof(vector_small), 255 + 1);
```

20162026:
Fix #123

### DIFF
--- a/include/beman/inplace_vector/inplace_vector.hpp
+++ b/include/beman/inplace_vector/inplace_vector.hpp
@@ -47,10 +47,10 @@ inline constexpr from_range_t from_range;
 // Smallest unsigned integer that can represent values in [0, N].
 template <size_t N>
 using smallest_size_t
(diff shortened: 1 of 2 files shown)


wusatosi (approved):
Ah! good catch, thanks.

jery4 (approved):

PR #152, opened 2026-09-16
use optional return types for fallible methods

redraincatching:
closes #119 by implementing [p3981](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3981r2.html), using `beman.optional` as specified in the issue.


include/beman/optional/detail/stl_interfaces/fwd.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+// Copyright (C) 2019 T. Zachary Laine
+//
+// Distributed under the Boost Software License, Version 1.0. (See
+// accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+#ifndef BEMAN_OPTIONAL_DETAIL_STL_INTERFACES_FWD_HPP
+#define BEMAN_OPTIONAL_DETAIL_STL_INTERFACES_FWD_HPP
+
+#include <beman/optional/detail/stl_interfaces/config.hpp>
```
JeffGarland, line 11: it's weird that there's a detail include from optional here....

include/beman/optional/detail/stl_interfaces/fwd.hpp
```diff
@@ -0,0 +1,85 @@
+// include/beman/optional/detail/stl_interfaces/fwd.hpp               -*-C++-*-
```
JeffGarland, line 1: copy pasta?

include/beman/optional/detail/stl_interfaces/fwd.hpp
JeffGarland: so I think the issue here is we're introducing a dependence on beman optional -- is that right?

redraincatching: yes, the original issue suggested using beman optional so i decided to go along with that
there's every chance i've mishandled bringing int the dependency

